PR #26, opened 2016-03-11
Add example of multiple cookbooks, no berkshelf

sean-horn:


jamescott:
@sean-horn what do you mean by "(Not intended Test-Kitchen behavior)"?

jamescott:
This is now updated. Thanks!

PR #31, opened 2016-03-21
Because a customer asked

pburkholder:
Words from the README at https://github.com/chef/mixlib-shellout/blob/master/README.md


jamescott:
https://docs.chef.io/ruby.html#shelling-out

Is updated!

PR #40, opened 2016-04-01
Updating logging instructions

dpetzel:


jamescott:
This is updated. I made the part about the Ohai run more neutral, as most of the time it's in the context of the chef-client run, but not 100%. Hope that works! Thanks for the PR.

james

jamescott:
This is fixed. Thank you!

PR #42, opened 2016-04-06
Make oc_id application examples safer

alexpop:
https://github.com/chef/chef-compliance/issues/739#issuecomment-206233353 describes why this docs change is needed.

It will prevent multiple oc_id['application'] blocks in chef-server.rb from stepping on each other.


alexpop:
merged

PR #44, opened 2016-04-07
Minor grammatical fixes and URL fix

fevrin:
Just some minor grammatical issues I came across while referencing the documentation.  Also, I think it makes sense to update the URL of the navbar link to GitHub so it points to the correct, current repo.


jamescott:
These are fixed. Thank you for the feedback!

james

PR #46, opened 2016-04-11
Fix attribute for start handlers

StephenKing:
The chef-client cookbook offers an attribute `node['chef_client']['config']['start_handlers']` to add start handlers.

The documentation, however, has an example that mentions `node['chef_client']['start_handlers']`. Correct this.


jamescott:
This is updated. Thanks!

PR #49, opened 2016-04-15
Fixed missing type, consistent capitalization

nmcspadden:
In the launchd docs, there's inconsistent capitalization for start_calendar_interval. In some examples, the keys are "Hourly", "Weekday", etc., and in later examples, they're lower case.  To be consistent with Chef styling, this PR makes them all lower case.

Additionally, the syntax section is missing the 'type' property, which is defined later on in the list of properties but not mentioned anywhere else (and is important for anyone distinguishing launch daemons from launch agents!).


jamescott:
Good catch. This is rebuilding right now. Thank you!

james

PR #55, opened 2016-04-28
Fix Chef client 12.9 link

obazoud:


jamescott:
Thanks -- I caught this independently the other day, so already good to go. But great catch!

james

PR #58, opened 2016-05-06
Make Windows omnitruck examples consistent.

rhass:
The Windows command line examples differed on the same rendered page in the way they were presented. This unifies the examples and makes them very clear as to what actually needs to be typed.


sersut:
👍

rhass:
Looks like this change was done elsewhere. Closing.

PR #69, opened 2016-06-08
adding a reference to the install_msi_url config setting for kitchen.yml

mwrock:


davidwrede:
@mwrock, this was merged in with a previous patchset.

PR #75, opened 2016-06-20
Updating documentation for chef-client options

PrajaktaPurohit:
REF: https://chefio.atlassian.net/browse/FLOW-417

On testing noticed that all flags other than -j get respected on subsequent chef-client runs with -i and -o specified together.

/cc @mwrock @ksubrama @lamont-granquist

### DIFF
--- a/includes_ctl_chef_client/includes_ctl_chef_client_options.rst
+++ b/includes_ctl_chef_client/includes_ctl_chef_client_options.rst
@@ -58,7 +58,7 @@ This command has the following options:
    |interval| |chef_client intervals| Default value: ``1800``.
 
 ``-j PATH``, ``--json-attributes PATH``
-   |json attributes|
+   |json attributes| Used to setup the first client run. For all the future runs with option -i the attributes are expected to be persisted in the chef-server.
 
    **Run-lists**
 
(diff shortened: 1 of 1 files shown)


mwrock:
:+1:  Love it when we can update docs!

lamont-granquist:
👍 

in case of future reference -- 

we can't change the behavior of -j with -i for the users of -o because it'll have a horrible effect on the non-override-run-list users.

what it would mean is that if you ran chef-client -j dna.json -i 3600 on a server that dna.json would override whatever was pulled from the chef-server -- effectively making those attributes read-only until someone logged in an killed the interval-running chef-client and removed the -j option.

pretty sure that there was an old, old bug we fixed on this in the tickets.opscode.com system, so this behavior is deliberate.

the only possible code change we could make here is to apply -j every time when running under -i, bug only if -o is specified, which is very far down the road of digging a deep hole in soft dirt with insufficient bracing...

davidwrede:
Change has been merged in. 

https://docs.chef.io/ctl_chef_client.html#options

davidwrede:
@PrajaktaPurohit, I'm going to close this PR because the change is in the docs.

PR #77, opened 2016-06-29
Corrected command for Compliance configuration

russellseymour:
Working with a partner they had issues configuring the compliance server.

This is because the Compliance server was set to `restart` rather than `reconfigure`.  This PR is for that change.

Once we did this on the compliance server it started to work properly and we did not get a redirect to the internal DNS address for the Dashboard in Azure.


davidwrede:
Fix is in the pipeline and should be merged soon. Thanks @russellseymour!

PR #81, opened 2016-07-10
Correct paths in "chef generate app" example output

drags:
Quick fix for some example output that didn't match reality. Looks like the author used the output from `chef generate app grantmc` for the example command output, but the given example command and the tree block that follows were pulled from the output of `chef generate app chef-repo`


davidwrede:
Thanks for the PR, @drags! Getting this into our Automate pipeline now.

PR #84, opened 2016-07-15
add kitchen.yml documentation related to kitchen transports

mwrock:
I have gotten several comments from customers that the transport is woefully under documented. This does not fix the entire documentation story but puts it somewhere that is googlable.


jjasghar:
:shipit:

PR #91, opened 2016-07-24
Update knife node show docs for nested attributes

wedge-jarrad:
The docs previously indicated that nested attributes were not understood
by knife however on my v12.11.18 install a construct like ''knife node
show NODE_NAME -a kernel.machine'' does work.

I presume that this functionality was added to knife at some point, though
I don't know specifically when.

### DIFF
--- a/step_knife/step_knife_node_show_attribute.rst
+++ b/step_knife/step_knife_node_show_attribute.rst
@@ -8,4 +8,12 @@ To list a single node attribute:
 
    knife node show NODE_NAME -a ATTRIBUTE_NAME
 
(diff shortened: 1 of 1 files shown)


glensc:
can confirm it works as low as chef client 11.16.4

davidwrede:
Thanks @wedge-jarrad and @glensc. I'll update the docs within the next day or two with this change.

glensc:
and i recall the `-a foo.bar` worked ever since i started using chef in 2013. i can check from git tags that it was 11.4.0 or so

davidwrede:
@wedge-jarrad, I squashed your PR commit into https://github.com/chef/chef-web-docs/commit/6796a65becd9a97fff473b614f9bf89ab5cee9e8#diff-12e3f1a35237b8726e32c46ef436f352. Your change is making it's way through our workflow pipeline and should be live in a couple hours.

Thanks for the catch!

- Dave

PR #93, opened 2016-07-29
Update step_resource_template_inline_method.rst

glensc:
fix example of inline method arguments. i'm pretty sure the variable not string was intended there.


jamescott:
This is updated. Thank you!

PR #100, opened 2016-08-11
Update url for ftp_proxy example

dustyburwell:


davidwrede:
Thanks for the fix, @dustyburwell!

PR #101, opened 2016-08-12
Fix the syntax for a referencing nested json keys in a couple inspec …

itmustbejj:
…examples


davidwrede:
Hey @itmustbejj, thanks for the syntax fixes!

PR #106, opened 2016-08-31
Promote cmp usage as it provides results with octal mode

alexpop:
To keep it in sync with the latest inspec docs:
https://github.com/chef/inspec/pull/999


jamescott:
is updated

PR #116, opened 2016-09-09
Fix typo in FIPS mode instructions

jerryaldrichiii:
This removes the extra `should` from [this](https://docs.chef.io/knife_bootstrap.html#fips-mode) document:

```
Should should only be enabled for environments that require FIPS 140-2 compliance
```

becomes

```
Should only be enabled for environments that require FIPS 140-2 compliance
```


jamescott:
Thank you @jerryaldrichiii ...

PR #121, opened 2016-09-30
Berkshelf isn't scary

nathenharvey:
This note makes Berkshelf seem scary and out-of-date.  It is neither of
those so the note should be removed.

Signed-off-by: Nathen Harvey nharvey@chef.io


tas50:
This should go

davidwrede:
Commit has been merged. Will be live in a couple hours.

davidwrede:
Change is live.

PR #127, opened 2016-10-03
5 digit modes don't need to be strings

tas50:


davidwrede:
Your commit has been merged. Change will go out in next doc push.

davidwrede:
Change is live.

PR #128, opened 2016-10-04
Chef requires Ruby 2.2.2+ now

tas50:
As of 12.14 we require Ruby 2.2.2


davidwrede:
@tas50, if people are running older versions of the client, will this version of Ruby break them? I'm wondering if we need to add a note to that bullet point.

tas50:
@davidwrede  We required 2.1 before so this had been wrong for a LONG time

tas50:
Also if they use 2.2+ they can still install say 12.12 or 12.13.

davidwrede:
Understood, @tas50. Thanks for the clarification.

PR #131, opened 2016-10-05
Cj/server rb

cjohannsen81:
As customers struggle to set the value right (api_fqdn) this has to be added. If there´s an equal sign the setting will be ignored and causes confusion, especially as ruby allow to set a value using an equal sign and we don´t do a syntax check which will be a better solution in the future.

Cheers
Christian


davidwrede:
@cjohannsen81, I tweaked your note and included it in the commit reference in this conversation, so I'm going to close out this PR and you can consider the change merged.

Thanks,
Dave

PR #133, opened 2016-10-07
Simplify how we present provides in custom resources

tas50:


lamont-granquist:
:+1:

also looks like we do document the precedence order of the provider resolver

lamont-granquist:
although if you look at the lists they run backwards, one or the other should be reversed.  or probably the first list should just be dropped.  this code looks good and is clearly better than what it replaces, but in context the examples seem to run backwards and don't agree with the text above it.

PR #134, opened 2016-10-09
Readme update to reflect the current state of Makefile

artem-sidorenko:
- `make master` instead of `make release`
- removal of dead upload target


davidwrede:
@artem-sidorenko, my apologies for not getting back to you sooner. We have made some big changes to the existing repo structure and have included updated instructions in the README as part of that. 

The Makefile was changed as part of these structure updates, and it will change again with another set up of upcoming fixes that we're planning to make. 

If you see a need for further improvement, feel free to submit another PR on the docs and thanks for helping to try and clean up the doc building experience.

PR #135, opened 2016-10-09
Ignore the virtualenvwrapper settings

artem-sidorenko:
its described in README this way, but the .venv is not in the gitignore


davidwrede:
Thanks for this change! It's been incorporated, so I'm closing out this PR.

PR #147, opened 2016-10-13
Remove metadata that we suggest people not use

tas50:
We have a lot of metadata that doesn't do anything and often confuses users. We even have Foodcritic rules that suggest people not use some of these. We should stop documenting it so that new users will stop using them.


lamont-granquist:
yeah, we might include a note somewhere that mentions they shouldn't be used, but this is better than documenting them like they actually do something...  👍

tas50:
Deprecation notice in Chef client would be nice

lamont-granquist:
its not really deprecated though its never been implemented, which is different...

we probably should go back and implement one or two of them...

PR #149, opened 2016-10-13
Link to Foodcritic.io instead of trying to list all the rules

tas50:
Foodcritic's site is up to date and includes nice examples. There's no need to duplicate work here.


davidwrede:
Closing this PR because the change is now in the live docs. https://docs.chef.io/foodcritic.html#rules

PR #151, opened 2016-10-13
Improve custom resources docs Round 1

tas50:
This is the first round of improvements to the custom resource documentation. It's mostly focused around providing a really simple first example of a custom resource so we don't scare people off. The 2nd round of improvements will focus around providing an example that uses pure Ruby and some of the more advanced concepts.

### DIFF
--- a/includes_custom_resources/includes_custom_resources.rst
+++ b/includes_custom_resources/includes_custom_resources.rst
@@ -3,12 +3,10 @@
 .. This file is hooked into a slide deck
 
 
-A custom resource:
+Custom Resources are:
 
-* Is a simple extension of |chef|
-* Is implemented as part of a cookbook
-* Follows easy, repeatable syntax patterns
-* Effectively leverages resources that are built into |chef|
-* Is reusable in the same way as resources that are built into |chef|
+A method of extending of |chef| to create your own custom resources that help to automate common patterns used in your cookbooks.
 
-For example, |chef| includes built-in resources to manage files, packages, templates, and services, but it does not include a resource that manages websites.
+Custom resources can:
+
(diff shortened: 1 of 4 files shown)


includes_custom_resources/includes_custom_resources_syntax.rst
```diff
@@ -5,25 +5,20 @@
 A custom resource is defined as a |ruby| file and is located in a cookbook's ``/resources`` directory. This file
 
 * Declares the properties of the custom resource
-* Loads current properties, if the resource already exists
 * Defines each action the custom resource may take
 
-The syntax for a custom resource is. For example:
+The syntax for a basic custom resource utilizing built-in chef resources is. For example:
```
thommay: I think we should make it clear that you can use ruby here, and it's safe to do so directly and that chef will take care of running it at the correct time.

includes_custom_resources/includes_custom_resources_syntax.rst
```diff
@@ -5,25 +5,20 @@
...
    end
 
    action :name do
-    # a mix of built-in Chef resources and Ruby
+    # built-in Chef resources
    end
 
-where the first action listed is the default action.
\ No newline at end of file
+where the first action listed is the default action.
```
thommay: I kind of think we should just teach people to be explicit about the default action.

includes_custom_resources/includes_custom_resources.rst
```diff
@@ -3,12 +3,10 @@
...
 
-A custom resource:
+Custom Resources are:
 
-* Is a simple extension of |chef|
-* Is implemented as part of a cookbook
-* Follows easy, repeatable syntax patterns
-* Effectively leverages resources that are built into |chef|
-* Is reusable in the same way as resources that are built into |chef|
+A method of extending of |chef| to create your own custom resources that help to automate common patterns used in your cookbooks.
```
thommay: Lose the second "of". This definition seems rather circular too.

thommay (changes requested):

tas50:
Closing this out to fix it in smaller bits

PR #172, opened 2016-10-24
Validators

MarkGibbons:
I tried to add the documentation for the resource validator parameters.  This was my first try at a documentation update and I'm sorry I don't understand the structure as well as I should.  

I think it's pretty confusing not to be able to find the old validators in the custom resource documentation.  I can't tell if chef plans to drop them, but until then I would like to be able to find them.

Thanks,
  Mark


snapshots/includes_dsl_custom_resource/includes_dsl_custom_resource_method_property_attribute_validation_parameter.rst
```diff
@@ -0,0 +1,113 @@
...
+
+Some examples of combining validation parameters:
+
+.. code-block:: ruby
+
+   property :spool_name, String, regex: /$\w+/
+
+.. code-block:: ruby
+
+   property :enabled, equal_to: [true, false, 'true', 'false'], default: true
```
Tensibai: Seems it is missing the Ruby type of the property before the validator part.

snapshots/includes_dsl_custom_resource/includes_dsl_custom_resource_method_property_attribute_validation_parameter.rst
```diff
@@ -0,0 +1,113 @@
+.. The contents of this file may be included in multiple topics (using the includes directive).
+.. The contents of this file should be modified in a way that preserves its ability to appear in multiple topics.
+
+
+A validation parameter is used to add zero (or more) validation parameters to an property.
```
Tensibai: Maybe `to a property` ? (unsure)

Tensibai (commented):
Few small points

MarkGibbons:
@Tensibai Thank you.  Should be fixed and tighter now.  Also got rid of the included file and created a new tagged section for the validator parameters.

chef_master/source/custom_resources.rst
```diff
@@ -857,7 +980,7 @@ For example, the following properties define ``username`` and ``password`` prope
 .. end_tag
 
 Block Arguments
-+++++++++++++++++++++++++++++++++++++++++++++++++++++
+-----------------------------------------------------
```
Tensibai, line 983: Why a different header size here ? This would break the overall presentation, or did I miss something ?

chef_master/source/custom_resources.rst
```diff
@@ -817,6 +818,128 @@ For example, the following properties define ``username`` and ``password`` prope
...
+
+Some examples of combining validation parameters:
+
+.. code-block:: ruby
+
+   property :spool_name, String, regex: /$\w+/
+
+.. code-block:: ruby
+
+   property :enabled, equal_to: [true, false, 'true', 'false'], default: true
```
Tensibai, line 939: Still missing a ruby type here, shouldn't it be `property :enabled, [TrueClass, FalseClass, String], equal_to: [true, false, 'true', 'false'], default: true` According to the above data (or if the type can be omitted, this worth a note even if I think it's a bad idea to not specify things)

snapshots/includes_dsl_custom_resource/includes_dsl_custom_resource_method_property.rst
```diff
@@ -6,13 +6,14 @@ Use the ``property`` method to define properties for the custom resource. The sy
...
+   property :name, ruby_type, default: 'value', validators: 'values'
 
 where
 
 * ``:name`` is the name of the property
-* ``ruby_type`` is the Ruby type, such as ``String``, ``Integer``, ``TrueClass``, or ``FalseClass``
-* ``default: 'value'`` is the default value loaded into the resource
+* ``ruby_type`` is the |ruby| type, such as ``String``, ``Integer``, ``TrueClass``, or ``FalseClass``
+* ``default: 'value'`` is the optional default value loaded into the resource
+* ``validators: 'values'`` are optional validation parameters and values
```
Tensibai: Worth keeping the `parameter`  name here for coherence with the rest of the doc.

Tensibai (commented):
The two first comments are valid for each file. Maybe worth using parameter_name instead to be clear it is not a syntax like `property :test_name, String, parameters: equal_to: [true,false]`

snapshots/includes_dsl_custom_resource/includes_dsl_custom_resource_method_property.rst
MarkGibbons: Wrapped this section in .. tag dsl_custom_resource_method_property to keep it coherent.

chef_master/source/custom_resources.rst
MarkGibbons: Updated the property description to note that ruby_type is optional.

chef_master/source/custom_resources.rst
MarkGibbons: It seemed to me that Block Arguments described something about properties that was useful in load_current_value processing and so should have the same header as property_is_set? and reset_property.  Grouping Block Arguments under the property section was confusing.  I'm willing to change this.

Tensibai: Fair enough :)

MarkGibbons:
@Tensibai Thank you again for another review.

davidwrede:
gif-keyboard-17449552733145657989

Thanks for getting this added back in! 

I will check with our engineers to see if the content for param validators should be deprecated. If so, I will add a note in the appropriate sections.

PR #178, opened 2016-10-26
Fix typo describing node.normal vs. node.default

GregMefford:
This sentence seemed to conflict with the paragraph above it.
I think this change is what was intended, though I'm not personally familiar with the difference between `node.normal` and `node.default`.

Signed-off-by: Greg Mefford greg@gregmefford.com


davidwrede:
gif-keyboard-7651794669826790660

PR #181, opened 2016-10-26
Remove Ohai plugins that are built in now

tas50:
One of these is mine. No one should ever get pointed at it now.


davidwrede:
gif-keyboard-7466506087067359546

PR #189, opened 2016-10-28
Give license and user key example for automate install

alexpop:


davidwrede:
gif-keyboard-15653732972861718479

PR #190, opened 2016-10-28
Update config_rb_client.rst - automatic proxy config

stefanwb:
Proxy settings already configured in ENV take precedence over those in the config file.
This follows POSIX standard.


davidwrede:
gif-keyboard-14153655697993102047

PR #192, opened 2016-10-28
added dsc_resource Windows Service example

drahamim:
Added an example of using dsc_resource to create a windows service.


davidwrede:
gif-keyboard-5265495415575269392

PR #197, opened 2016-11-01
Inline include files for archived releases

colincam:
This PR inlines all include files and converts absolute URIs to relative to make the archived releases independent of each of and of the current doc set.


davidwrede:
gif-keyboard-10175861923464961578

davidwrede:
Great work, @colincam!

PR #198, opened 2016-11-02
[IPO-475] adding docs on Elasticsearch and Kibana authentication

adamleff:
This change adds documentation for the Elasticsearch and Kibana authentication which will be included in an upcoming release.

The doc is being excluded from robot indexing and the left-hand nav until the feature is available in a stable release.


davidwrede:
gif-keyboard-3002850093257627021

PR #202, opened 2016-11-03
Putting the external Elasticsearch at the bottom and making it optional

yvovandoorn:
By having Elasticsearch at the top of the document, it makes it as if the external dep is required for Automate -- but it isn't. 

I propose moving it to the bottom and making it optional.


davidwrede:
gif-keyboard-1194949568723974715

PR #204, opened 2016-11-03
Fix knife data bag from file example

rneu31:


davidwrede:
gif-keyboard-11538558921892296377

PR #210, opened 2016-11-06
Simplify custom resource warning text

tas50:

### DIFF
--- a/chef_master/source/custom_resources.rst
+++ b/chef_master/source/custom_resources.rst
@@ -3,7 +3,7 @@ Custom Resources
 =====================================================
 `[edit on GitHub] <https://github.com/chef/chef-web-docs/blob/master/chef_master/source/custom_resources.rst>`__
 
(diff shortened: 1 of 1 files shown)


MarkGibbons:
Hi @tas50 
The link https://docs.chef.io/custom_resources_notes.html is removed with this pull request.  That page has important clues for how to write 12.5 custom resources that are not documented anywhere else in chef-web-docs that I can find.  Where to put methods related to an action was pretty baffling without that doc.  I'm not even sure the note is correct but using action_class does work.

tas50:
I'm going to get a PR cut to move some of this logic into the main doc. Having two places to look seems really confusing and I think we can clean a lot of this up

davidwrede:
@tas50, I'll merge this PR, but as an FYI, we will be removing the version picker and all `/release/*` content from the site soon. This means we should start flattening and including content that works for all minor versions of a given product/project release (i.e. all client 12.x content should be included in the current client docs until we cut client 13, and then all of the client 12.x content gets placed on our new archive site).

davidwrede:
gif-keyboard-4549736293514790787

PR #216, opened 2016-11-07
Correct node wording

colincam:


colincam:
Not a real change, just used for demo purposes

PR #143, opened 2016-10-11
Move recipe DSL and custom resource pages

tas50:
Nuke the resources page under Extension APIs
Move custom resources under Cookbook Reference
Move recipe DSL under Cookbook Reference -> Recipes

Signed-off-by: Tim Smith tsmith@chef.io

PR #218, opened 2016-11-07
Make Recipe DSL reachable in the nav.

RichardBronosky:
This solution makes Recipe DSL a sibling of Recipe rather than a child.
The heirarchy is altered, but the fine grain links are retained. This
appears to have been davidwrede's intent according to
https://github.com/chef/chef-web-docs/pull/143

NOTE: To make such a significant heirarchy change I had to lint the
JSON. This resulted in a lot of minor whitespace and comma corrections.
I wasn't trying to be pedantic.

### DIFF
--- a/chef_master/source/_templates/nav-docs.html
+++ b/chef_master/source/_templates/nav-docs.html
@@ -41,26 +41,26 @@
         "title": "Packages &amp; Platforms",
         "hasSubItems": true,
         "subItems": [
-           {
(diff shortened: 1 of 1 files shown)


colincam:
A corresponding change to index.rst needs to added to this commit in order to keep left nav and site map consistent.

davidwrede:
gif-keyboard-9825845433568645139

PR #223, opened 2016-11-08
Remove kind_of: from the resource property definitions

MarkGibbons:
In 12.5 custom resources the property type is positional we don't need to use kind_of.


tas50:
👍

davidwrede:
gif-keyboard-4827036941095415266

PR #231, opened 2016-11-11
Chef Client deprecations documentation

thommay:
This creates the base deprecations page and a single detailed
deprecation for discussion.

cc @tas50

### DIFF
--- a/chef_master/source/chef_deprecations_client.rst
+++ b/chef_master/source/chef_deprecations_client.rst
@@ -3,14 +3,7 @@ Chef Deprecation Warnings
 =====================================================
 `[edit on GitHub] <https://github.com/chef/chef-web-docs/blob/master/chef_master/source/chef_deprecations_client.rst>`__
 
-.. tag client_deprecations
-
-When we wish to remove a feature or an API in Chef, we try to first mark it with a deprecation warning that contains a link to a description of the change and how to fix it.
-
-.. end_tag
-
-Example
-=====================================================
+When we wish to remove a feature or an API in Chef, we try to first mark it with a deprecation warning that contains a link to a description of the change and how to fix it. For example:
 
 .. code-block:: ruby
 
@@ -34,42 +27,54 @@ All Deprecations
 =====================================================
 
 .. list-table::
-  :widths: 50 350 50
+  :widths: 50 230 40 80
   :header-rows: 1
 
   * - ID
     - Description
-    - Version
+    - Deprecated
+    - Expected Removal
   * - :doc:`CHEF-1 </deprecations_json_auto_inflate>`
     - Consumers of JSON are now required to be explicit in how it is turned in to a Chef object.
-    - 12.7.2  
+    - 12.7.2
+    - 13.0  
   * - :doc:`CHEF-2 </deprecations_exit_code>`
     - Chef's exit codes are now defined so that it's easy to understand why Chef exited.
     - 12.11.4
+    - 13.0
   * - :doc:`CHEF-3 </deprecations_chef_gem_compile_time>`
     - When using the ``chef_gem`` resource, the phase to install the gem in must be specified.
     - 12.1.0
+    - 13.0
   * - :doc:`CHEF-4 </deprecations_attributes>`
     - Various improvements have been made to attribute syntax.
     - various
+    - varies
   * - :doc:`CHEF-5 </deprecations_custom_resource_cleanups>`
     - Various improvements have been made to custom resource syntax.
     - various
+    - varies
   * - :doc:`CHEF-6 </deprecations_easy_install>`
     - The ``easy_install`` resource will be removed.
     - 12.10.17
+    - 13.0
   * - :doc:`CHEF-7 </deprecations_verify_file>`
     - The ``verify`` metaproperty's ``file`` substitution will be removed.
     - 12.5.1
+    - 13.0
   * - :doc:`CHEF-8 </deprecations_supports_property>`
     - The ``supports`` metaproperty will be removed.
     - 12.14.17
+    - 13.0
   * - :doc:`CHEF-9 </deprecations_chef_rest>`
     - The ``Chef::REST`` API will be removed.
     - 12.7.2
+    - 13.0
   * - :doc:`CHEF-3694 </deprecations_resource_cloning>`
     - Resource Cloning will no longer work.
     - 10.18.0
+    - 13.0
   * - :doc:`CHEF-0 </deprecations_internal_api>`
     - Many internal APIs have been improved.
     - various
+    - varies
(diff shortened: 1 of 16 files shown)


lamont-granquist (approved):

tas50:
Very cool to see. My only input is that I'd love for us to come up with a nice title that conveys the forward looking nature of the doc without making it seem like some sort of binding deprecation contract.

chef_master/source/chef_deprecations_client.rst
```diff
@@ -0,0 +1,45 @@
...
+.. code-block:: ruby
+
+   Deprecated features used!
+     JSON auto inflation is not supported (CHEF-1) at (irb):7:in `irb_binding'.
+     Please see https://docs.chef.io/chef-client/deprecations/json_auto_inflate.html for further details and information on how to correct this problem.
+
+Fixing Deprecations
+=====================================================
+
+To test your code for deprecations, you can put Test Kitchen in a mode where any deprecations cause the chef run to fail. Ensure your ``.kitchen.yml`` includes:
```
tas50, line 17: We should note that this requires Test Kitchen 1.13 or later

chef_master/source/chef_deprecations_client.rst
```diff
@@ -0,0 +1,45 @@
...
+All Deprecations
+=====================================================
+
+.. list-table::
+  :widths: 50 350 50
+  :header-rows: 1
+
+  * - ID
+    - Description
+    - Version
```
tas50: Version is a pretty vague column name. This is the version we introduced this deprecation in correct?

chef_master/source/deprecations_json_auto_inflate.rst
```diff
@@ -0,0 +1,35 @@
...
+=====================================================
+`[edit on GitHub] <https://github.com/chef/chef-web-docs/blob/master/chef_master/source/deprecations_json_auto_inflate.rst>`__
+
+.. tag deprecations_json_auto_inflate
+
+Internally, the ``Chef::REST`` class attempts to guess which Chef class a JSON document relates too, and then automatically turns the JSON in to that class.
+
+.. end_tag
+
+This deprecation was added in Chef 12.7.2, and will be removed permanently in Chef 13.
```
tas50: This should be clarified. We deprecated this _functionality_ and the _functionality_ will be removed in Chef 13.

chef_master/source/chef_deprecations_client.rst
```diff
@@ -0,0 +1,45 @@
...
+  * - :doc: `CHEF-1 </deprecations_json_auto_inflate>`
```
tas50: We need a status column as well. That allows us to list things as either Pending Full Deprecation or Deprecated in Chef XYZ.

thommay: I don't think I agree. Things in this document should only be there once they're deprecated, at which point we know which version they are deprecated in.

chef_master/source/chef_deprecations_client.rst
thommay: Yes. Suggestions welcome to how we can word this better without blowing out the table.

chef_master/source/chef_deprecations_client.rst
tas50: I think we're using deprecated two different ways, and by that I mean I used it wrong in in the comment above. 

We know which chef version we deprecated a feature in, but we also need to list the version we actually changed the behavior in so a status column is important. When a person looks at this page in 2 years they need to see both functionality that's already changed (in Chef 13, 14, whatever) and functionality that will change in the future. The ability for the document to not only list upcoming compat breaks, but also past compat breaks allows it to be useful in scoping the work involved in client upgrades.

thommay: OK, so `deprecated` and `removed in` as two separate column heads?

| name | deprecated in | removed in |
| --- | --- | --- |
| foo | 12.14 | 13 |
| bar | 10.18 | 14 |

davidwrede:
Hey @thommay, I updated the table to add in an "Expected Removal" column. I believe a column like this is important because while it's nice to know when a deprecation notice was introduced, it's just as important to know, at a glance, when the feature/flag/thing will be removed from the product. 

Might not be applicable in all cases and if you think of a better column, feel free to add it.

thommay:
@lamont-granquist @tas50 I'd really appreciate re-review of this, especially the custom_resource section.

chef_master/source/deprecations_attributes.rst
```diff
@@ -0,0 +1,65 @@
...
+-------------
+
+Convert method syntax to bracket syntax by using brackets to denote attribute names. The code below is identical in function to the example above:
+
+.. code-block:: ruby
+
+  node['chef']['server'] = "https://my.chef.server"
+  chef_server = node['chef']['server']
+
+Set and Set_Unless
```
stevendanna, line 41: Not sure the best place to track this, but once we have a "persistent data" API, it would be good to revisit the examples in this section.

chef_master/source/deprecations_json_auto_inflate.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Remediation
+=====================================================
+
+You now need to explicitly create a new object of the desired type.
+
+.. code-block:: ruby
+
+  name = "my_environment"
+  Chef::Environment.from_hash chef_server_rest.get("environments/#{name}")
```
stevendanna, line 34: Alternatively: `Chef::Environment.load(name)` which is a bit cleaner.

thommay: I wanted to demonstrate getting the JSON and then transforming it. Not sure if there's a better way.

chef_master/source/deprecations_custom_resource_cleanups.rst
```diff
@@ -0,0 +1,141 @@
...
+    updated = true
+  end
+
+Remediation
+------------
+
+.. code-block:: ruby
+
+  action :foo do
+    updated_by_last_action true
```
stevendanna, line 115: Is this the best remediation in most cases?  Isn't it better to restructure the code to use `converge_by` blocks if possible?  Is that worth mentioning here?

stevendanna (approved):
Overall these look good to me.  I've left some small questions but none are blocking.

chef_master/source/deprecations_custom_resource_cleanups.rst
thommay: My feeling is that we should provide the most minimal thing that works. It will be hard to demonstrate rewriting into `converge_by` blocks succinctly and I worry that if the remediation is too hard people simply won't bother.

thommay:
@davidwrede I think this is ready to go.

tas50 (approved):
We'll tweak this in the future, but we should get it out there.

davidwrede:
gif-keyboard-11937980919000671279

PR #232, opened 2016-11-11
Specify location of delivery.rb file

jsmartt:
The doc page that describes the configuration options for a file should probably include the location of that file (if it's not obvious). It wasn't obvious to me, so hopefully this helps others.


davidwrede:
gif-keyboard-16577271758449166448

PR #239, opened 2016-11-14
Improve description of chef-solo

tas50:
Chef-client is always open source. This was really confusing. It's just
a different way to executing chef-client, which we should be more clear
about. Also lets make sure people understand the benefits of
client/server better.


davidwrede:
gif-keyboard-13298107064098073827

PR #243, opened 2016-11-15
Rename Supermarket API page + cleanup

tas50:
This appears to be a find / replace job on the old community site API
docs. It's a bit confusing as is. This improves it a bit by using the
Supermarket name in more places and removing some of the "community"
workding that really meant community site and not community cookbook.


davidwrede:
gif-keyboard-16204781357086774041

PR #246, opened 2016-11-15
Update URLs to avoid redirects and use https

tas50:


davidwrede:
gif-keyboard-7136081179576363898

PR #247, opened 2016-11-15
Remove `:` from `action` on group resource docs

smith:
`action` isn't a symbol, so it should be `action` instead of `:action`.


davidwrede:
gif-keyboard-16400194452213037905

PR #249, opened 2016-11-15
Fix case on Ohai client.rb docs

drenalin23:
According to the online docs, something like Ohai.plugin_path can be used in client.rb.  After a lot of searching and poking around I discovered the following:

In Ohai 8.6.0+ you can use the new format for the ohai attributes in client.rb
They are actually lower case ohai, in various testing I found that Ohai.plugin_path does not work but ohai.plugin_path does (and then discovered/confirmed that usage in the chef-client community cookbook).

This updates the docs in chef_master.  The CONTRIBUTING doc doesn't go into much detail, but I was guessing that updates in chef_master directory path will get auto-generated into the snapshots directory?

### DIFF
--- a/chef_master/source/config_rb_client.rst
+++ b/chef_master/source/config_rb_client.rst
@@ -382,15 +382,15 @@ Ohai configuration settings can be added to the client.rb file.
 
 .. tag config_rb_ohai_settings
 
-``Ohai.directory``
+``ohai.directory``
    The directory in which Ohai plugins are located.
 
-``Ohai.disabled_plugins``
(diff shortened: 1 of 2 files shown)


thommay (approved):

tas50 (approved):

davidwrede:
Thanks @drenalin23 for this! Apologies for the delay in getting this checked in. And don't worry about updating the other doc sets in the snapshots directory. That dir will be going away shortly.

gif-keyboard-2172374817579542186

PR #250, opened 2016-11-15
[IPO-492] adding docs for the visibility node delete cmds

adamleff:
Two new commands are available in Automate 0.6.0 for deleting nodes
from Visibility.


davidwrede:
In addition to formatting, I also added in some of the latest commands that weren't reflect in the left nav.

davidwrede:
gif-keyboard-13369638460519391285

PR #251, opened 2016-11-16
[IPO-490] add nav links to ES auth docs

adamleff:
This feature is launching with Automate 0.6.0, so we're surfacing it!


davidwrede:
gif-keyboard-14089071096321964804

PR #253, opened 2016-11-16
recommend python SimpleHTTPServer

thommay:
We've already got python installed, so let's use the http server that's built in.


davidwrede:
gif-keyboard-8197503811250875240

PR #255, opened 2016-11-16
chef_gem is only a package in the loosest sense

thommay:
The normal "use the package resource" note makes very little sense for `chef_gem`


tas50:
+1

davidwrede:
gif-keyboard-6187688654432526514

PR #259, opened 2016-11-17
Update Automate backup docs to include digest, encryption, and retries

ryancragun:


davidwrede:
gif-keyboard-13815866787879495116

PR #264, opened 2016-11-18
Add note about hard-to-find automate-ctl install-build-node error

srenatus:
This is meant to make it smoother for the next person who hits this problem.

Could have gone into the _"About Proxies"_ section, too, I suppose, but I think it is more closely tied to the build-node/runner installation.


davidwrede:
@sr, looks good. At some point I think this message should be moved to the troubleshooting_automate topic (along with possible remedies).

PR #266, opened 2016-11-18
config.rb change - no_proxy does not honor IP ranges

jeremiahsnapp:
most HTTP clients including curl, wget and ruby's net/http do not honor IP ranges in the no_proxy setting


davidwrede:
gif-keyboard-1228241892405946416

PR #271, opened 2016-11-20
Remove duplicate section from cookbook_versions

wedge-jarrad:
The listing of cookbook version operators is duplicated (copy/paste
error?). Delete it.


davidwrede:
gif-keyboard-6021452910243402730

PR #273, opened 2016-11-21
Fix a few links and update the versions

alexpop:


davidwrede:
gif-keyboard-6192631260155299876

PR #278, opened 2016-11-25
Need to include all of the supporting trusted_cert files

mattray:
The file from nginx does not work on my AWS-deployed self-signed cert. I needed all of the supporting files.


davidwrede:
gif-keyboard-16773549558982019930

PR #279, opened 2016-11-25
Fix sentence structure and add highlighting

usrme:
Noticed something was off when reading the documentation in terms of the structure of the edited sentence and the missing code highlight from `knife bootstrap` as it refers directly to a sub-command.


davidwrede:
gif-keyboard-17650998140352692613

PR #281, opened 2016-11-29
Minor tweaks to the getting started page

tas50:
There's no need to talk about chef-zero with new users. It's a backend
tech that enables cool tooling, but itself isn't the cool tooling. We
should also remove references to definitions since no one new to Chef
needs to know about these. Finally ChefDK comes with Cookstyle, but
"Policy" isn't really a tool. It's a Chef feature and it seems out of
place in this list.


davidwrede:
gif-keyboard-4940497702963816112

PR #284, opened 2016-12-01
Fix typos for data_collector params in client client

adamleff:


davidwrede:
gif-keyboard-13661974733786825745

PR #285, opened 2016-12-01
Update Chef Server prereq docs for ipv4 installs

sean-horn:
Chef Servers intended to run in ipv4 mode should not have any references to ipv6 addresses in their /etc/hosts file.

Clarify this requirement in the prereqs section.
Without this prereq, ipv4 mode Chef Server install/reconfigure fails because postgres will not be available at 127.0.0.1:5432 during the reconfiguration.


davidwrede:
gif-keyboard-9709479984256782027

PR #286, opened 2016-12-01
Runners only run build node jobs

sean-horn:
Clarify that runners only run build node jobs in Automate 0.6.7

Push Jobs must still be used for workflow-triggered infrastructure node runs and delivery-truck/delivery-sugar deploys.


chef_master/source/job_dispatch.rst
```diff
@@ -19,6 +19,18 @@ Chef Automate release 0.6.0 includes a new SSH-based job execution layer. Under
...
+
+   By default, Job Dispatch nodes cannot be used to deploy builds on infrastructure nodes. Infrastructure nodes are usually used as part of testing, but may not be present in your environment. Infrastructure nodes are not build nodes. Currently, as of Automate 0.6.7, Job Dispatch can only be used directly against build nodes.
+
+   * One typical environment would be an Acceptance environment on which both automated and manual testing of new code might happen against a production-like environment. The deploy previous to automated testing will be triggered by a build node triggering a Push Jobs client on the infrastructure node. This will cause the infrastructure node's chef-client to run on the infrastructure node, thus deploying the build. The delivery-truck cookbook's deploy recipe in https://github.com/chef-cookbooks/delivery-truck/blob/master/recipes/deploy.rb#L34 currently assumes that a Push Jobs client will be available on infrastructure nodes for the deploy phase. This behavior is the default, but can be customized to potentially trigger the chef-client run through means other than a Push Jobs job, such as Job Dispatch.
```
dmccown: `such as Job Dispatch` you may want to replace this with `such as ssh`. I believe that's another approach that may be taken, but it would not have any ties in the with the Job Dispatch system itself.

dmccown (dismissed):

chef_master/source/job_dispatch.rst
```diff
@@ -19,6 +19,18 @@ Chef Automate release 0.6.0 includes a new SSH-based job execution layer. Under
 - View the queue of jobs in the Chef Automate UI
 - View and test runner status via the Chef Automate UI
 
+Known Issues
+--------------
+
+.. warning:: If you need to run windows build and infrastructure nodes, stop now. Job Dispatch is currently incompatible with Windows. If you are running only linux build/infrastructure nodes, please continue.
```
dmccown: This messaging isn't totally correct. You can use windows infrastructure nodes (as long as they can run a push client). The job dispatch feature uses `runners` and those are for linux at the moment (ubuntu, redhat, centos).

PR #289, opened 2016-12-02
Move Test Driven Infrastructure into general chef books

tas50:
This book is 3 years old and is getting a bit dated. It's still helpful,
but we shouldn't feature it as the definitive guide for new users. A
third edition would be super helpful.

### DIFF
--- a/chef_master/source/kitchen.rst
+++ b/chef_master/source/kitchen.rst
@@ -313,6 +313,4 @@ For more information about test-driven development and Kitchen:
 
 * `kitchen.ci <http://kitchen.ci>`_
 * `Getting Started with Kitchen <http://kitchen.ci/docs/getting-started/>`_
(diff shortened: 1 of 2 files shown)


chef_master/source/kitchen.rst
```diff
@@ -313,6 +313,4 @@ For more information about test-driven development and Kitchen:
 
 * `kitchen.ci <http://kitchen.ci>`_
 * `Getting Started with Kitchen <http://kitchen.ci/docs/getting-started/>`_
-* `Test-Driven Infrastructure with Chef, 2nd Edition <http://shop.oreilly.com/product/0636920030973.do>`_, by Stephen Nelson-Smith (O'Reilly Media)
 * `How Can I Combine Berks and Local Cookbooks? <https://coderwall.com/p/j72egw/organise-your-site-cookbooks-with-berkshelf-and-this-trick>`_
```
thommay, line 316: I'm not sure this coderwall link is all that, either. It seems to be linked all over the place.

thommay (approved):
I think we should clobber that coderwall link but either way 👍

davidwrede:
gif-keyboard-6322505244873109494

PR #298, opened 2016-12-05
fix references to chef_oauth2_app_id

robbkidd:
zuul

There is no `chef_oauth2_app`, only `chef_oauth2_app_id`.


davidwrede:
gif-keyboard-13655618198279824202

PR #299, opened 2016-12-05
fix references to sample private Supermarket URL

robbkidd:
[Living in your own private Supermarket](https://youtu.be/n7t7cGwN7_0).


davidwrede:
gif-keyboard-4589573461324744063

PR #300, opened 2016-12-05
Add search caveat for encrypted data bags

stevendanna:
This occasionally causes users confusion. See chef/chef-server#876


stevendanna:
The wording of this is a bit awkward.  I'm open to alternatives.  The point is: "search doesn't work with encrypted data bags"

davidwrede:
gif-keyboard-18314889904533922286

PR #301, opened 2016-12-06
Minor build improvements

stevendanna:
- Avoid double testing of PR branches (PRs will still be tested, just
  once instead of twice)

- Move `clean` down in the make file so it isn't the default target


davidwrede:
gif-keyboard-6061780885448378105

PR #303, opened 2016-12-06
Adding deprecations 11 and 12

adamleff:
* 11: property names that conflict with existing methods are disallowed
 * 12: the launchd `hash` property is renamed due to deprecation 11

Related to chef/chef#5606

/cc @thommay @lamont-granquist

### DIFF
--- a/chef_master/source/chef_deprecations_client.rst
+++ b/chef_master/source/chef_deprecations_client.rst
@@ -77,6 +77,14 @@ All Deprecations
   * - :doc:`CHEF-10 </deprecations_dnf_package_allow_downgrade>`
     - DNF package provider and resource do not require ``--allow-downgrade`` anymore.
     - 12.18.31
+    -
+  * - :doc:`CHEF-11 </deprecations_property_name_collision>`
+    - An exception will be raised if a resource property conflicts with an already-existing property or method. 
+    - 12.19.x
+    - 
+  * - :doc:`CHEF-12 </deprecations_launchd_hash_property>`
+    - An exception will be raised whenever the ``hash`` property in the launchd resource is used.
+    - 12.19.x
     - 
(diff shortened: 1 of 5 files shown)


davidwrede:
Holding until #5606 goes out with the next stable build.

thommay:
I think that deprecations should get pushed out as soon as possible.

davidwrede:
Entries are needed in the root deprecations table. I'll get those added in. Deprecation warnings (and any doc changes for that matter) go out with the corresponding stable release, so we'll hold on this PR until that happens.

thommay:
Not sure that's the right thing to do here; we're shipping chef clients that contain those URLs in the current channel, and have certainly encouraged users to use current to test whether bugs are fixed, etc. So there are some (admittedly small) proportion of users that are seeing those URLs right now.

lamont-granquist:
It also adds more churn to the release process, and was blocking the release pending the URLs going live.

The only downside might be user confusion over live URLs with deprecation notices that haven't gone out yet, but really anyone surfing through the deprecations list should be able to figure that out, particularly given that we've got the index of chef versions.  I've seen lots of confused users over the years, but it would take a particularly badly confused user to be confused over prematurely live deprecation docs.

adamleff:
Add my :+1: to releasing this now. If this were a new feature, I'd hold off on releasing the docs until the feature is released to quell disappointment, etc. But since these are deprecations, I'm in favor of mashing the button.

davidwrede:
Because these are deprecations, I agree that it can't hurt anything to release the docs now, so I'll get the changes merged in; however, we clearly state on the [downloads page](https://downloads.chef.io/chef/current/12.18.51) that Chef docs only cover stable releases, so please keep that in mind if you have tests for current releases that are gated on doc URLs being live.

PR #304, opened 2016-12-06
Update docs for launchd resource

adamleff:
* `hash` property has been renamed to `plist_hash`
 * Keys for `start_calendar_interval` need to start with a capital
   letter, at least in macOS Sierra

Related to chef/chef#5606

/cc @thommay @lamont-granquist


davidwrede:
Holding until #5606 goes out with the next stable build.

tas50:
@davidwrede This shipped. We should ship the doc change

davidwrede:
@tas50 This change didn't make it out with the last stable release (12.18.31). This #5606 was merged in after that, so when 12.19 goes out, then we'll get this doc fix merged in.

PR #312, opened 2016-12-08
Rearrange Chef Server Docs

sean-horn:
- Move the nav around to make more sense for Chef Server. It should be one layer deeper Chef -> Chef Server -> Managing the server

- Add Upgrade HA link for better discovery


sean-horn:
I'll figure this out later.

PR #313, opened 2016-12-08
Doc changes for windows_package resource 'returns' property

Vasu1105:
Doc changes after verifying the 'returns' property of windows_package resource accepts integer and array of integers.


davidwrede:
gif-keyboard-1228241892405946416

PR #310, opened 2016-12-07
Add deprecation of Chef::Platform methods

lamont-granquist:
Signed-off-by: Lamont Granquist <lamont@scriptkiddie.org>

PR #321, opened 2016-12-12
Revert "Add deprecation of Chef::Platform methods"

davidwrede:
Reverting this because the content is tied to an upcoming client release (not a current client release).

Reverts chef/chef-web-docs#310


davidwrede:
Decided to keep this change in. Versioning information in the content should keep current users (i.e. those running < 12.18) from assuming it applies to them. That and I won't be reverting a revert commit (and mudding up the history) later on.

PR #328, opened 2016-12-14
Kitchen -> Test Kitchen

tas50:
Test Kitchen is the proper name


davidwrede:
gif-keyboard-10781139541991800868

PR #332, opened 2016-12-14
package -> gem in the chef_gem docs

tas50:
We kept calling it a package when it's a gem. It's just a tad bit confusing


thommay (approved):

davidwrede:
gif-keyboard-2141211568834446966

PR #348, opened 2016-12-16
chef_gem doesn’t support multi package installs

tas50:
Don’t tell people you can pass an array here


davidwrede:
gif-keyboard-8468439503659629999

PR #360, opened 2016-12-22
Remove incomplete HWRP docs

tas50:
1) We highly recommend against writing what some people call the HWRP
2) These docs were incomplete and did more harm than good

We need to rework what is included in this doc, but I think removing items that aren’t helpful is a good first step.


cheeseplus:
This is definitely better than providing more confusing information.

PR #366, opened 2016-12-29
Fix option description string

briancurt:


davidwrede:
gif-keyboard-11789507261350161844

PR #379, opened 2017-01-05
Remove deprecated ctl commands

blakestier:
`list-applications` and `delete-application` were part of the github v1 integration. That integration is deprecated as are these commands.


davidwrede:
gif-keyboard-14637824366049466296

PR #381, opened 2017-01-05
Update release notes for prior ChefDK releases

tas50:


tas50:
@davidwrede here's the previous gem updates you asked about before xmas break

kagarmoe:
Thanks!

PR #383, opened 2017-01-06
Adding new lint tests for docs

davidwrede:
Added two new linting tests:

 - lint_edit_check verifies that the `edit on GitHub` link is near the top of every source file.
 - lint_nav_check reports on which source files are not in the left nav and index (sitemap) files. 

Out of the two, only lint_edit_check is run as part of the build process. At a later point in time, we may add the nav check in there as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ build/
 
 /.delivery/cli.toml
 
+/.vscode
+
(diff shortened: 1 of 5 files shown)


davidwrede:
Merging changes in now, review will come later.

PR #392, opened 2017-01-10
Fix #391

### LINKED ISSUE #391: Remove `cookbooks_path` reference from kitchen examples
This was added in https://github.com/chef/chef-web-docs/pull/26 but reviewing test-kitchen code this was never add, though there is an old PR https://github.com/test-kitchen/test-kitchen/pull/393. This functionality doesn't exist at all in the code and should be removed.

cheeseplus:
`cookbooks_path` never existed and doesn't work so it should be nuked from orbit

Signed-off-by: Seth Thomas <sthomas@chef.io>


cheeseplus:
Dupe of #388

PR #396, opened 2017-01-13
Updated dsc_resource resource document to reflect the module_version support added for the resource.

Aliasgar16:
Done.

### DIFF
--- a/chef_master/source/release_notes.rst
+++ b/chef_master/source/release_notes.rst
@@ -4340,7 +4340,7 @@ The **dsc_resource** resource allows any DSC resource to be used in a Chef recip
 
 Syntax
 +++++++++++++++++++++++++++++++++++++++++++++++++++++
-.. tag resource_dsc_resource_syntax
+.. tag resource_dsc_resource_syntax_12_2
 
 A **dsc_resource** resource block allows DSC resourcs to be used in a Chef recipe. For example, the DSC ``Archive`` resource:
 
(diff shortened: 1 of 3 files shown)


Aliasgar16:
@davidwrede I am not sure that the change that I have made in `release_notes.rst` is correct. I have made this change just to make the `dtags` consistent. The change is in the section `What's New in 12.2`. However the change made to the `dsc_resource` will be released in the upcoming versions of `Chef` and so it does not pertain to `Chef 12.2` version, but to make the `dtags` consistent I had to make this change. Please let me know what can be done here.

Also this PR is dependent on the [PR](https://github.com/chef/chef/pull/5701) where the actual functionality is implemented.

davidwrede:
Thanks @Aliasgar16 for the updates and you're right. We don't want to retroactively change previous release notes just because the content is shared in two different places. 

My fix will be to append a _12_2 to the shared block tag in the 12.2 rel notes and remove the new property. These changes will make it clear that when we refer to that version of dsc_resource, we're clear about what functionality existed at that time.

I'll also add in a "New in Chef client 12.x" in the description of the property. Any time a new property/flag/resource/whatever is introduced, changed, deprecated, or removed, this sentence will get updated. And once the stable release number is known for this code change, I'll update the note.

So, no additional action needed on your part. We'll handle this clean up work.

Aliasgar16:
Thanks @davidwrede

btm:
https://github.com/chef/chef/pull/5701 was just merged so it will be in the February Chef release.

PR #401, opened 2017-01-18
Clarify use of --json-attributes

zts:
Someone in the #chef IRC channel was confused about use of this option - although they'd read this documentation, they mistook the example as being something that would work for setting override_attributes.  While the docs are accurate, I agree it's easy to misinterpret them if you're not paying close attention.


davidwrede:
gif-keyboard-9076463741067357915

PR #410, opened 2017-01-23
Changes data bag to Chef vault.

kagarmoe:


afiune (approved):
I am not sure if the Markdown is rendering correctly for the first section where I put a list, It looks like it will be displayed on the same line. 

```
* There are two mechanisms to manage secrets in Chef Automate:
  - Using an encrypted data bag
  - Using chef vaults
```
Out of that, thank you for changing this!! 

:shipit:

PR #424, opened 2017-01-27
Edit data_collector token reference

larryebaum:
Clarify and be explicit that even the token value (even if default is used) must be added to chef server config...
Signed-off-by: Larry Eichenbaum <larryebaum@chef.io>


davidwrede:
gif-keyboard-2083133696089539098

PR #435, opened 2017-01-31
Adding back in apt cookbook reference to docs-builder cookbook

davidwrede:
The apt cookbook is referenced in the delivery kitchen erb template file. Adding it back in and bumping the version again so that our docs build. We can investigate doing a more permanent fix later on.


kagarmoe:
gif-keyboard-5119194698538278038

PR #439, opened 2017-02-02
Correct partial search request method.

dolvany:
A small edit to correct the request method indicated for api partial search functionality.


davidwrede:
gif-keyboard-7230814050909767367

PR #448, opened 2017-02-10
Update Chef Server API POST /users body

sean-horn:
Add an example Chef Server API POST body for /users that works


kagarmoe:
gif-keyboard-12162865951660561014

PR #449, opened 2017-02-10
Clarify default solver used by Berkshelf

ncs-msmith:
The default solver selection behavior is a little more nuanced than the documentation currently describes.


kagarmoe:
gif-keyboard-6346447494100748124

PR #450, opened 2017-02-10
Use double quotes to expand variable.

azusa:
To expand a variable a in string literal (in ruby), they have to use double quotes instead of single quotes.
Thus I cannot get the template '#{node.chef_environment}.erb' in my cookbook.


kagarmoe:
gif-keyboard-14671801753195101316

PR #455, opened 2017-02-14
Updates Solaris platforms

kagarmoe:


chef_master/source/platforms.rst
```diff
@@ -48,7 +48,7 @@ The following table lists the Foundational platforms for the chef-client and oha
      - ``5``, ``6``, ``7``
    * - Solaris
      - ``sparc``, ``x86``
-     - ``10``, ``11``
+     - ``10``, ``10 1\13`` (``10 U11``), ``11.0``, ``11.2``, ``11.3``
```
scotthain: Not sure how the formatting ends up with rst docs, but the end result for these should read something like:
```
10 1/13 ("U11"), 11.2, 11.3
```

scotthain (dismissed):
Let me know if you have any questions regarding these!

chef_master/source/platforms.rst
kagarmoe: Yep.  To clarify, are we supporting 11.0?

chef_master/source/platforms.rst
```diff
@@ -48,7 +48,7 @@ The following table lists the Foundational platforms for the chef-client and oha
      - ``5``, ``6``, ``7``
    * - Solaris
      - ``sparc``, ``x86``
-     - ``10``, ``10 1\13`` (``10 U11``), ``11.0``, ``11.2``, ``11.3`` 
+     - ``10``, ``10 1\13`` (``"10U11"``), ``11.2``, ``11.3``
```
scotthain: I think there's an extra `10` in here

PR #457, opened 2017-02-14
Correct config.json "build_cookbook" key

mattlqx:
From what I can tell, the correct key is `build_cookbook` as that is what is used in all the examples.


kagarmoe:
gif-keyboard-9175916255355337149

PR #458, opened 2017-02-14
Remove redundant step in Example Project App

mattlqx:
This is the same thing as the step immediately above it.


kagarmoe:
gif-keyboard-15707944840284483767

PR #466, opened 2017-02-15
Used {...} instead of do...end in not_if example

jdgoins:
Per Ruby style guide.


kagarmoe:
gif-keyboard-7650931170773197298

PR #471, opened 2017-02-17
Deprecate Cisco platforms post 12.19 release

scotthain:

12.19 will be the final release where we natively build and test Cisco platforms.


davidwrede:
gif-keyboard-1592210085816581818

PR #472, opened 2017-02-21
Updated community.rst

packtpartner:
Added a newly published book to "Recommended reading"


kagarmoe:
Thanks for the suggestion.  We will take a look at the book you've recommended and include it if we think it meets the needs of our community.

packtpartner:
That's great @kagarmoe . Will be looking forward to hearing back from you once you've had a look at the book.

PR #475, opened 2017-02-21
fix a typo in ruby type of mount resource's supports

t2y:
See https://github.com/chef/chef/blob/master/lib/chef/resource/mount.rb#L45

### DIFF
--- a/chef_master/source/resource_mount.rst
+++ b/chef_master/source/resource_mount.rst
@@ -41,7 +41,7 @@ The full syntax for all of the properties that are available to the **mount** re
      password                   String
      provider                   Chef::Provider::Mount
      subscribes                 # see description
-     supports                   Array
+     supports                   Hash # defaults to { :remount => false }
      username                   String
      action                     Symbol # defaults to :mount if not specified
    end
@@ -239,9 +239,9 @@ This resource has the following properties:
    .. end_tag
 
 ``supports``
-   **Ruby Type:** Array
+   **Ruby Type:** Hash
 
-   An array of options for supported mount features. Default value: ``:remount``.
+   Specify a Hash of supported mount features. Default value: ``remount: false``.
 
 ``username``
    **Ruby Type:** String
--- a/chef_master/source/resource_mount.rst
+++ b/chef_master/source/resource_mount.rst
@@ -41,7 +41,8 @@ The full syntax for all of the properties that are available to the **mount** re
      password                   String
      provider                   Chef::Provider::Mount
      subscribes                 # see description
-     supports                   Hash # defaults to { :remount => false }
+     supports                   Hash # defaults to { :remount => false } (preferred)
+                                Array # defaults to { :remount => true } (non-preferred)
      username                   String
      action                     Symbol # defaults to :mount if not specified
    end


chef_master/source/resource_mount.rst
```diff
@@ -239,9 +239,9 @@ This resource has the following properties:
    .. end_tag
 
 ``supports``
-   **Ruby Type:** Array
+   **Ruby Type:** Hash
 
-   An array of options for supported mount features. Default value: ``:remount``.
+   Specify a Hash of supported mount features. Default value: ``remount: true``.
```
burtlo: Looking at the code and the comment above the default value, when not specifying a Hash of values, is `remount: false`.

chef_master/source/resource_mount.rst
```diff
@@ -41,7 +41,7 @@ The full syntax for all of the properties that are available to the **mount** re
      password                   String
      provider                   Chef::Provider::Mount
      subscribes                 # see description
-     supports                   Array
+     supports                   Hash # defaults to { :remount => false }
```
burtlo: The interesting thing about the code is that `supports` does in fact support an `Array` as well. When an `Array` is specified the resulting `Hash` are all of the keys defined in the `Array` with values set to `true`.

burtlo: This mount resource using a `Hash`:
```
mount '/tmp' do
  supports { remount: true }
end
```
is equivalent to using an `Array`
```
mount '/tmp' do
  supports [ :remount ]
end
```

t2y: Hmm, it's true, but it's not appropriate.

chef_master/source/resource_mount.rst
t2y: Oh, I fixed.

chef_master/source/resource_mount.rst
burtlo: I wanted to add some comments in here around the code based on a conversation I was having someone at Chef. Hash all you want. Array all you want. 😄 I like seeing how code works and talk about it.

burtlo:
I think this is good change to the docs. 👍 

Also, it took me a moment: the link that you provided above now points to a different line of code. Did you mean: https://github.com/chef/chef/blob/987bcb5dd740637693dbb6723e8730f32da43696/lib/chef/resource/mount.rb#L34-L35?

t2y:
> Also, it took me a moment: the link that you provided above now points to a different line of code. Did you mean: https://github.com/chef/chef/blob/987bcb5dd740637693dbb6723e8730f32da43696/lib/chef/resource/mount.rb#L34-L35?

Yes, exactly.

t2y:
This is the source code at those days when I sent this PR.

* https://github.com/chef/chef/blob/937faa7e42c6df03b1deec0e8bbc785b7ba33e90/lib/chef/resource/mount.rb#L45

t2y:
@burtlo @kagarmoe The review of my changes has finished. I think there's no task for me. Is it right?

kagarmoe:
gif-keyboard-4234956650618166531

PR #491, opened 2017-02-28
Add Ohai Deprecation docs

tas50:


tas50:
We can probably improve these, but they're a starting point

tas50:
Thanks for the edits. I'll have some updates to all this eventually, but it would be great if we could get this shipped out ASAP

PR #497, opened 2017-03-03
Update quick_start.rst

mrjcleaver:
Presumably you don't suggest changing your "system Ruby" (as used by the rest - system - of MacOS X), only the Ruby used by the user.

It'd be nice to suggest a Ruby Version Manager also.


kagarmoe:
gif-keyboard-17966507520681616078

PR #499, opened 2017-03-07
Update setup_visibility_chef_automate.rst

charlesjohnson:
This change brings clarity to the documentation, by making the intent of the feature more obvious. There are very few use cases where people have a chef server, but would still prefer for nodes to send data directly to the Automate server.


davidwrede:
gif-keyboard-5712121053725010588

PR #500, opened 2017-03-07
Update chef_repo to reflect valid chefignore locations

tduffield:
chefignore files can technically go anywhere in a chef-repo at any level. The most common are either at the root of the chef-repo or in a cookbook itself.


kagarmoe:
gif-keyboard-13458799943594458568

PR #504, opened 2017-03-08
the partial_search method is gone

thommay:


thommay:
This has been deprecated for absolutely ages and will be gone entirely in 13.

davidwrede:
gif-keyboard-9146959592342297230

PR #506, opened 2017-03-08
Fix apt_update syntax

tas50:
Fix the syntax to be correct


davidwrede:
gif-keyboard-10781139541991800868

PR #507, opened 2017-03-08
Fix attribute merge details

adamleff:
The current docs are incorrect regarding merge behavior when an
attribute value is a hash. Correcting the docs to match the
actual behavior.


kagarmoe:
Thanks!

PR #508, opened 2017-03-09
Update secondary and untested platforms

tas50:
Move openSUSE 13 out of secondary since it's EOL now and replace it with 42 (leap) which we test extensively via cookbook integration testing. Remove OS X 10.9 from chefDK as it's not on the download site and note that Windows 2008 is not supported there as well.


davidwrede:
gif-keyboard-9790552469634267530

PR #511, opened 2017-03-10
Merge web branch commits into master branch

davidwrede:


davidwrede:
This also contains commits that update our doc deployments and streamlines our publishing process.

PR #521, opened 2017-03-15
Fixes issue with openssl command

dmccown:
Only the server certificate was being rendered with the command given.
The command was changed to output the entire chain and gives
instructions on how to pull the needed parts into a file.

Signed-off-by: David McCown <dmccown@chef.io>


kagarmoe:
gif-keyboard-8577844256497293804

PR #529, opened 2017-03-21
clarify systemd_unit documentation

rmoriz:
- tune example: add action, use shortcut to write more readable unit files inside recipe
- clarify, that the name property must include the type (or `systemd-analyze verify` will fail`) 
- explain, which verification command is used (`systemd-analyze verify`)


kagarmoe:
gif-keyboard-4401272116997588273

PR #534, opened 2017-03-25
Update Ruby types in custom resources

tas50:
Fixnum has been deprecated in favor of Integer in Ruby
NilClass, TrueClass, and FalseClass are confusing and aren't necessary. You can just use true/false/nil


davidwrede:
gif-keyboard-2537292052764723733

PR #535, opened 2017-03-25
Remove duplicate 12.5 warning

tas50:
Custom resource themselves are entirely new for 12.5. There's no need to call out this one small change between LWRPs and custom resources.


davidwrede:
gif-keyboard-14488549827822945977

PR #536, opened 2017-03-25
Use Ruby 1.9+ hash syntax

tas50:
Hash rockets were deprecated in 2007 when Ruby 1.9 was released


davidwrede:
gif-keyboard-5271232399969352694

PR #543, opened 2017-03-29
Fix formatting

shoekstra:


davidwrede:
travisCI is down atm, so verified linting and builds on my local machine.

PR #546, opened 2017-03-30
Fix two small issues

srenatus:


kagarmoe:
gif-keyboard-3562453431250124634

PR #548, opened 2017-03-30
remove references to provider-internal resource-state properties of systemd_unit

nathwill:
enabled, active, static, masked are internal state-properties that shouldn't be user-facing (the user-provided values are ignored and overwritten by the provider). this updates the docs to remove misleading references to these properties, and instead demonstrates the intended use of the corresponding actions.


davidwrede:
gif-keyboard-15514193461313690103

PR #556, opened 2017-04-04
Correct elasticsearch port ranges

stevendanna:
ES actually listens on the first available port in the 9200-9300 range
for the client endpoint and 9300-9400 for the inter-node connections.


davidwrede:
gif-keyboard-15508784878837904144

PR #562, opened 2017-04-05
Kick Automate for job type

sean-horn:
Sometimes, Automate is confused about the job-type to builder/runner resource availability mapping.
You can kick it to get it processing jobs again.


kagarmoe:
gif-keyboard-12516742019718153474

PR #571, opened 2017-04-06
Libvirt will be deprecated in Chef 14 not 13

tas50:
Missed the window here


kagarmoe:
gif-keyboard-8530496337868481140

PR #575, opened 2017-04-09
Update berkshelf.rst

mrjcleaver:
Re: using Policyfiles instead: "This is unless you are using Chef Automate, which for now can only use Berkshelf."


davidwrede:
gif-keyboard-1857421907452304847

PR #577, opened 2017-04-10
Update windows.rst

cixelsyd:
The command to schedule a task is incorrect for windows 2008R2.  I believe the command is faulty for most modern windows OS.  The suggested command works on windows 2008R2 x86_64.


kagarmoe:
gif-keyboard-18429899408934470956

PR #580, opened 2017-04-10
Expand ext elastic

thomascate:
We've had some confusion from customers about setting up External ES clusters since the only example is https with no custom port specified. I added an example of multi-host, with non-ssl and the default ES port to cover some ground.


davidwrede:
gif-keyboard-7659904143303551481

PR #583, opened 2017-04-10
Remove attributes from metadata.

iennae:
Attributes hasn't been removed (yet), but is not advised for current use.
Use of attributes creates immediate technical debt because it becomes
outdated as the cookbook evolves. To use it requires folks to update
information in three places within the code itself, the documentation
in the README, and the metadata. When it gets uploaded to be shared
it's converted into JSON which isn't human readable so using attributes
is not helpful. The newest foodcritic includes it as a warning if it's used.


kagarmoe:
image

PR #585, opened 2017-04-12
Remove duplicate notes about 12.14

martinisoft:
This is a quick pass to clean up what looks to be 3 similar notes about 12.14 and 12.5 changes. I've taken a little liberty with the re-wording here and tried to keep the same tone of the original document.


kagarmoe:
gif-keyboard-1545898846818547766

Good catch!

PR #588, opened 2017-04-12
Update SLES information for appropriate projects

scotthain:

@davidwrede


scotthain:
We are now building on these platforms but we haven't released any of the versions that include SLES builds. The next official releases will include SLES packages.

davidwrede:
gif-keyboard-6044828054794073281

PR #596, opened 2017-04-17
Improve the initial custom resource example

tas50:
This current example is overly complex and not actually necessary. If
someone submitted that to one of our cookbooks we'd kick it back. We
should start people off with a simple example of a pure chef resources
based custom resoure. That sort of custom resource requires just the
actions to be defined and doesn't need state loading, which is an
advanced topic.


davidwrede:
gif-keyboard-8403065115667208647

PR #599, opened 2017-04-19
Fix invalid long_description ruby code

szymonpk:
It is invalid and it wouldn't work. New line can be placed after `(`, not before.


davidwrede:
gif-keyboard-5158513278730573385

PR #600, opened 2017-04-19
Added a missing word.

jshvrsn:
If I'm understanding blacklists correctly, then there's a very important word missing in the docs.


kagarmoe:
gif-keyboard-1545898846818547766

PR #601, opened 2017-04-20
update sample URL

jaakla:
The sample with .php is very confusing. There is no point to download URL http://somesite.com/index.php (as html) and put it to /var/www/customers/public_html/index.php as source, unless you really do not know what you are doing. Maybe this example was meant to be a joke, but it is really just confusing and does not help to understand things properly.


davidwrede:
gif-keyboard-5278108743294518289

PR #602, opened 2017-04-20
Updating Supermarket info in Chef Automate install doc

davidwrede:
Minor note and tweak added to setup content so that people who want to use a private Supermarket don't have to potentially run `setup` twice.


kagarmoe:
gif-keyboard-17213110099492321173

PR #603, opened 2017-04-20
correct attribute references in supermarket.rb

robbkidd:
Fixes the missing `default` and name of the attributes in `supermarket.rb`. Also adds some S3 settings that were missing.


davidwrede:
gif-keyboard-8786420419324377045

PR #604, opened 2017-04-22
Describe the exact directory path of the file

daddygongon:
When doing step 5 of editing default.rb, 
because the location of the file is deep and uncertain,
I looked around to find the file for a while.
Thus the exact directory path will be described explicitly.


davidwrede:
gif-keyboard-5514514672524350356

PR #607, opened 2017-04-24
OS X -> macOS rename

davidwrede:
Updating name of Mac OS to be consistent with current naming.


tas50:
If only we still had those variables ;)

davidwrede:
Indeed, @tas50. 😄

PR #610, opened 2017-04-24
Properly note that path was deprecated in Chef 12 not 13

tas50:


kagarmoe:
gif-keyboard-10599663223008235891

PR #618, opened 2017-04-24
Remove an untruth about the .chef dir on a chef-repo

tas50:
You don't need these to work with a server. You need a knife.rb, but it can be your personal knife.rb file


davidwrede:
gif-keyboard-7651794669826790660

PR #619, opened 2017-04-25
Move around / dedupe node and chef-client pages

tas50:
The content on neither of these make sense. It’s all a mix of node and client type data. This clears up some of that and makes the docs a bit more distinct.


kagarmoe:
gif-keyboard-3686478377366320034

PR #626, opened 2017-04-26
Remove partial changelog and update CLI options

tas50:
The CLI options were missing a few things. I've also removed the changelog. The changelog for Foodcritic is pretty extensive and we just can't fit the list of changes in this way. It'll get entirely out of hand within a few months. We have a changelog for that.


davidwrede:
gif-keyboard-10010989575426586418

PR #631, opened 2017-05-01
Remove Chef 11 style default directory

tas50:
This directory structure isn't required with Chef 12+ and custom resources are Chef 12.5+ so there's no reason to show the old structure or even note it.

### DIFF
--- a/chef_master/source/custom_resources.rst
+++ b/chef_master/source/custom_resources.rst
@@ -370,7 +370,7 @@ httpd.conf.erb
      TypesConfig /etc/mime.types
    </IfModule>
 
-Copy it as shown, add it under ``/templates/default``, and then name the file ``httpd.conf.erb``.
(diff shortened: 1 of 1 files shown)


akitada:
There are other docs that note the `default` directory. Take https://docs.chef.io/resource_template.html for example.

tas50:
It should really go away there as well. There's no reason to push users towards this directory structure since we shipped Chef 12 in 2014

akitada:
`chef generate template httpd` generates `templates/httpd.erb` and **`templates/default`**.
You might want to change this, too.

tas50:
Fixed the generator here https://github.com/chef/chef-dk/pull/1246

davidwrede:
Created #633 to track removal across other topics.

davidwrede:
tenor-151024916

PR #634, opened 2017-05-02
Update chocolatey_package to clarify source arg

michaeltlombardi:
The source argument in the existing document may lead a naive user (like me!) to assume that the path _to the package_ must be specified, instead of the path _to the folder containing the package_. The error message for the former is the unhelpful `No candidate version available for <PackageName>`.

This also makes it clear that packages can be installed from UNC.


davidwrede:
tenor-4652250

michaeltlombardi:
Awesome, thanks for further clarifying and including this!

PR #637, opened 2017-05-04
Add important notice about postgres subnet config

sean-horn:
Add an extended explanation of why postgres subnet rules for frontend/backend access should be replicated identically throughout the cluster.

### DIFF
--- a/chef_master/source/install_server_ha.rst
+++ b/chef_master/source/install_server_ha.rst
@@ -326,6 +326,8 @@ which would result in the following two entries being added to the ``pg_hba.conf
 
 Running ``chef-backend-ctl reconfigure`` on all the backends will allow that frontend to complete its connection.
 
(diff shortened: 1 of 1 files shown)


chef_master/source/install_server_ha.rst
```diff
@@ -326,6 +326,8 @@ which would result in the following two entries being added to the ``pg_hba.conf
 
 Running ``chef-backend-ctl reconfigure`` on all the backends will allow that frontend to complete its connection.
 
+.. important:: All members of both the frontends and backend clusters MUST contain identical sets of the ``postgresql.md5_auth_cidr_addresses`` subnet rules set for all systems, frontend and backend. This guarantees that when a cluster change of state occurs, all members can still speak with one another as needed. Any other configuration is at risk of having unsuccessful failovers where followers are unable to communicate with the new leader, or the frontend cluster is unable to communicate with the new leader.
```
jeremiahsnapp, line 329: The `postgresql.md5_auth_cidr_addresses` setting is only applied on the backends. Not on the frontends. But the value set on the backends should contain the subnets of the frontends if the subnets are different from that of the backends' subnets.

davidwrede:
tenor-246082920

PR #643, opened 2017-05-05
Update notifier docs to include compliance and correct attribute name

kmacgugan:

PR #645, opened 2017-05-08
Last notification->notifier change in delivery.rb

ksubrama:


ksubrama:
@kmacgugan I missed one edit during the review of #643 .

Issue #206, opened 2016-11-04
Config option rubygems_url is not documented


joerg:
I just found an option in the chef code [1] about an option "rubygems_url" which if used for installing gems defined in the metadata.rb.
I could not find any documentation about this option but I think its quite handy.

[1] https://github.com/chef/chef/blob/master/lib/chef/cookbook/gem_installer.rb

jflemer-ndp:
@lamont-granquist looks like you added this feature (in chef/chef@1d063284f755b996a6563d7210496afd716a9f2d), and I am in need of using it. Docs would help.

iennae:
https://github.com/chef/chef-web-docs/pull/647

iennae:
This is now documented. Thanks for opening up the issue.

PR #647, opened 2017-05-08
Adding rubygems_url setting documentation for air gapped environments

iennae:
Resolves Issues 
#206
https://github.com/chef-cookbooks/rubygems/issues/15

### DIFF
--- a/chef_master/source/config_rb_client.rst
+++ b/chef_master/source/config_rb_client.rst
@@ -300,7 +300,7 @@ This configuration file has the following settings:
 ``role_path``
    The location in which role files are located. Default value: ``/var/chef/roles``.
 
-``ruby_gem_url``
+``rubygems_url``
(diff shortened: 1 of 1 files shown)


iennae:
@coderanger can you take a peek at this. Not sure if I should add any additional information with regards to air gapped environments or if that should be a separate document somewhere.

coderanger (approved):
This is good enough for the moment. Long term it should also be mentioned in a narrative section in https://github.com/chef/chef-web-docs/blob/master/chef_master/source/resource_chef_gem.rst because that's probably where people will come in looking for "how do I use chef_gem with my own gems server".

coderanger:
I am reminded while we're talking on Slack, this should mention the config option is new in 13.0.

chef_master/source/config_rb_client.rst
```diff
@@ -300,6 +300,11 @@ This configuration file has the following settings:
 ``role_path``
    The location in which role files are located. Default value: ``/var/chef/roles``.
 
+``ruby_gem_url``
```
coderanger: Should be `rubygems_url`.

davidwrede:
tenor-184674784

PR #649, opened 2017-05-09
Automate single enterprise only

sean-horn:
Automate supports a single enterprise only, as of Automate 0.7.239


davidwrede:
tenor-206347671

PR #653, opened 2017-05-11
Tiered updates

kevinreedy:
I ran through this document tonight, as a customer will be going through this installation tomorrow. I found a couple of small issues that I've updated here.


davidwrede:
tenor-148700807

PR #655, opened 2017-05-11
Updated deprecation/removed warning to the supports property.

iennae:
This is mentioned on the deprecation page but hasn't been updated for this page. I'm not sure what the right mechanism to label this property as deprecated is but I gave this a go. Correct it if this isn't appropriate. It could be deleted as well but that might not provide information to folks as to what happened if they look at this page first.

### DIFF
--- a/chef_master/source/resource_common.rst
+++ b/chef_master/source/resource_common.rst
@@ -79,11 +79,12 @@ The following properties are common to every resource:
    Ensure that sensitive resource data is not logged by the chef-client. Default value: ``false``. This property only applies to the **execute**, **file** and **template** resources.
 
 ``supports``
+   .. warning:: This property was deprecated in Chef 12.14 and removed in Chef 13.0.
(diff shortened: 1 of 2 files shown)


iennae:
@lamont-granquist and @coderanger could give additional feels about this if needed.

tas50:
@iennae That same block is found in another doc. Check the Travis failure.

kagarmoe:
It is nice to note things that are deprecated for users.  It can help reduce frustration during debugging.

kagarmoe:
tenor-189739794

lamont-granquist:
this table is supposed to cover that:

https://github.com/iennae/chef-web-docs/blob/230fd2f376fff4d43f08b355643f47f0809a3af0/chef_master/source/chef_deprecations_client.rst

tas50:
We should certainly link to the deprecation page directly

PR #659, opened 2017-05-13
Remove ohai cookbook documentation

tas50:
Not a single bit of this is true. We do need documentation for the ohai cookbook, but this isn't it.


davidwrede:
tenor-213783103

PR #656, opened 2017-05-12
Document the changes in https://github.com/berkshelf/berkshelf/pull/1690

coderanger:
New source-y things.

PR #660, opened 2017-05-14
Revert "Document the changes in https://github.com/berkshelf/berkshelf/pull/1690"

davidwrede:
Reverts chef/chef-web-docs#656


davidwrede:
I need to revert without reverting the other changes. The content will come out, just not via this mechanism.

PR #667, opened 2017-05-17
Remove unnecessary content from the chef overview page

tas50:
We should not be pointing brand new users to definitions and providers. There's no reason to even mention them in the getting started overview page. This removes a large amount of information, which rightfully makes Chef look simpler.


davidwrede (approved):
👍  on removing the clutter

davidwrede:
tenor-79408588

PR #665, opened 2017-05-17
Update config_yml_kitchen.rst

jmery:
Add documentation for provisioner setting `deprecations_as_errors`.

PR #668, opened 2017-05-17
Update config_yml_kitchen.rst  -  provisioner / deprecations_as_errors

notapatch:


davidwrede:
@notapatch Seems like dup of #665.

PR #675, opened 2017-05-19
Add kitchen-terraform

Maniacal:
New Context has a successful kitchen-terraform driver that I would like to get added.


kagarmoe:
tenor-23653810

PR #684, opened 2017-05-23
LCR link updates

tpetchel:


davidwrede (approved):
Spot checked the links and they seem to resolve correctly.

PR #687, opened 2017-05-25
Document CHEF-18 deprecation

coderanger:
Merge when ready, doesn't need to wait for the Chef code side.


chef_master/source/chef_deprecations_client.rst
```diff
@@ -94,6 +94,9 @@ All Deprecations
     - Deprecation of run_command
     - 12.18
     - 13.0
+  * - :doc:`CHEF-18 </deprecations_local_listen>`
+    - Deprecation of local mode listening.
+    - 13.1
```
davidwrede, line 99: Can you add in the expected removal info to this row entry? The build breaks because a fourth value isn't present in this row.

kagarmoe:
tenor-46795628

PR #691, opened 2017-05-29
[MSYS-582]updated knife bootstrap docs

dheerajd-msys:


jeremiahsnapp:
This looks great! The only suggestion I have is to change the following sentence in chef_master/source/ctl_automate_server.rst and chef_master/source/knife_bootstrap.rst.

```
All config files inside ``client.d`` directory that end in ``.rb`` are loaded and get copied into ``/etc/chef/client.rb``; other non-``.rb`` files are ignored.
```

As far as I can tell from the code at the following link _all_ files in the client.d directory get copied. Not just `.rb` files.

https://github.com/chef/chef/blob/v13.1.32/lib/chef/knife/core/bootstrap_context.rb#L248-L260

davidwrede:
tenor-158948784

Issue #582, opened 2017-04-10
add note to chef-server-ctl restore command


vinyar:
URL: https://docs.chef.io/server_backup_restore.html#id4

Note: Docs should explicitly note that `chef-server-ctl backup/restore` should not be used for migrating to a new version of chef server. Restores should only be done between identical versions.  If upgrade is desired, chef server should be upgraded after functionality is verified.

PR #702, opened 2017-06-05
do not use backup/restore across versions of chef server

brewn:


sean-horn:
👍 LGTM. :shipit:

brewn:
#582

PR #703, opened 2017-06-05
Chef Vault incompatible with old-style client cert

sean-horn:
Chef Vault does not work with old-style chef-client certs, only with current public/private keypairs


irvingpop:
Questions:   
- How old of a chef client would you need (era or major version) to get those?  
- How would a customer know that those keys exist in their environment?  (is there any way to proactively audit)

kagarmoe:
Do we have a documented workflow for updating an older system to use chef vault?  If I'm running a 2 year-old system (or older?) and throw in ``gem chef-vault``, how does it go for me?

kagarmoe:
tenor-195759238

PR #704, opened 2017-06-05
Clear up the text for converge_by

EasyAsABC123:
Importantly removed it would `touch the file` which is what doesn't happen during a dry run.
To make it clear that `converge_by` doesn't deal with idempotency and is instead useful for `why-run`.

Peer-documented @iennae
Review Requested to @lamont-granquist

### DIFF
--- a/chef_master/source/custom_resources_notes.rst
+++ b/chef_master/source/custom_resources_notes.rst
@@ -181,9 +181,9 @@ If you do need to write code which mutates the system through pure-Ruby then you
      end
    end
 
-The ``converge_by`` block gets why-run correct and will just touch "/tmp/foo" instead of actually doing it. The ``converge_by`` block is also responsible for setting ``update_by_last_action``.
+The ``converge_by`` block gets why-run correct and will print to ``Chef::Log.info`` "touch /tmp/foo" and not run ``::FileUtils.touch "/tmp/foo"``. The ``converge_by`` block is also responsible for setting ``update_by_last_action``.
 
-In order to use ``converge_by`` correctly you must ensure that you wrap the ``converge_by`` with an idempotency check otherwise your resource will be updated every time it is used and will always fire notifications on every run.
+The ``converge_by`` block does not deal with idempotency and will set the ``update_by_last_action`` to ``true`` everytime the block executes.  By wrapping this block with an idempotency check like ``converge_if_changed`` or ``unless File.exist?("/tmp/foo")`` this will stop ``update_by_last_action`` from being set.
 
 .. code-block:: ruby
 
--- a/chef_master/source/custom_resources_notes.rst
+++ b/chef_master/source/custom_resources_notes.rst
@@ -183,11 +183,9 @@ If you do need to write code which mutates the system through pure-Ruby then you
 
 When the ``converge_by`` block is run in why-run mode, it will only log ``touch "/tmp/foo"`` and will not run the code inside the block. 
 
-The ``converge_by`` block does not do any checking for idempotency and always sets ``updated_by_last_action``.  A
-``converge_by`` block that is not wrapped in an idempotency check will always cause the resource to be updated, and
-will always cause notifications to fire.  A properly written resource should wrap all ``converge_by`` checks with an
-idempotency check, or the [``converge_if_changed``](https://github.com/chef/chef-web-docs/blob/master/chef_master/source/custom_resources.rst#converge_if_changed) block should be used instead.   As the ``converge_if_changed`` API
-wraps a ``converge_by`` block with an idempotency check for you.
+A ``converge_by`` block that is not wrapped in an idempotency check will always cause the resource to be updated,
+and will always cause notifications to fire.  To prevent this, a properly written resource should wrap all
+``converge_by`` checks with an  idempotency check.  The [``converge_if_changed``](https://github.com/chef/chef-web-docs/blob/master/chef_master/source/custom_resources.rst#converge_if_changed) block may be used instead  which will wrap a ``converge_by`` block with an idempotency check for you.
 
 .. code-block:: ruby
 
--- a/chef_master/source/custom_resources_notes.rst
+++ b/chef_master/source/custom_resources_notes.rst
@@ -183,7 +183,11 @@ If you do need to write code which mutates the system through pure-Ruby then you
 
 When the ``converge_by`` block is run in why-run mode, it will only log ``touch "/tmp/foo"`` and will not run the code inside the block. 
 
-The ``converge_by`` block is also responsible for setting ``update_by_last_action``, however it does not deal with idempotency and will set the ``update_by_last_action`` to ``true`` everytime the block runs.  By wrapping this block with an idempotency check like ``converge_if_changed`` or ``unless File.exist?("/tmp/foo")`` this will stop ``update_by_last_action`` from being set.
+The ``converge_by`` block does not do any checking for idempotency and always sets ``updated_by_last_action``.  A
+``converge_by`` block that is not wrapped in an idempotency check will always cause the resource to be updated, and
+will always cause notifications to fire.  A properly written resource should wrap all ``converge_by`` checks with an
+idempotency check, or the ``converge_if_changed`` API should be used instead.   As the ``converge_if_changed`` API
+wraps a ``converge_by`` block with an idempotency check for you.
 
 .. code-block:: ruby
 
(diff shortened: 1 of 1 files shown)


chef_master/source/custom_resources_notes.rst
```diff
@@ -181,9 +181,9 @@ If you do need to write code which mutates the system through pure-Ruby then you
      end
    end
 
-The ``converge_by`` block gets why-run correct and will just touch "/tmp/foo" instead of actually doing it. The ``converge_by`` block is also responsible for setting ``update_by_last_action``.
+The ``converge_by`` block gets why-run correct and will print to ``Chef::Log.info`` "touch /tmp/foo" and not run ``::FileUtils.touch "/tmp/foo"``. The ``converge_by`` block is also responsible for setting ``update_by_last_action``.
```
lamont-granquist: This is awkward reading.

The prior text was incorrect and mangled somehow.

Maybe:

```
When the ``converge_by`` block is run in why-run mode, it will only log ``touch "/tmp/foo"`` and will not run the code inside the block.
```

I really dislike throwing the `Chef::Log.info` and `FileUtils.touch` details at the reader.  It is also likely better to be clear that the code inside the block is not executed -- there is no magic other than the block simply is not run.  Also "gets why-run correct" is awkward (i think I wrote that one initially, it sounds like me) and should really be changed.

lamont-granquist (changes requested):

EasyAsABC123:
@lamont-granquist agreed, let me know if the new text makes more sense :) i only added the `converge_if_changed` part as i thought it was close to the same, i now know i am wrong.

chef_master/source/custom_resources_notes.rst
```diff
@@ -181,9 +181,9 @@ If you do need to write code which mutates the system through pure-Ruby then you
      end
    end
 
-The ``converge_by`` block gets why-run correct and will just touch "/tmp/foo" instead of actually doing it. The ``converge_by`` block is also responsible for setting ``update_by_last_action``.
+When the ``converge_by`` block is run in why-run mode, it will only log ``touch "/tmp/foo"`` and will not run the code inside the block.
```
lamont-granquist, line 184: oops I didn't mean for the sentence "The ``converge_by`` block is also responsible for setting ``update_by_last_action``." to get dropped as well, that should probably get re-added...

chef_master/source/custom_resources_notes.rst
```diff
@@ -181,9 +181,9 @@ If you do need to write code which mutates the system through pure-Ruby then you
      end
    end
 
-The ``converge_by`` block gets why-run correct and will print to ``Chef::Log.info`` "touch /tmp/foo" and not run ``::FileUtils.touch "/tmp/foo"``. The ``converge_by`` block is also responsible for setting ``update_by_last_action``.
+When the ``converge_by`` block is run in why-run mode, it will only log ``touch "/tmp/foo"`` and will not run the code inside the block. 
 
-The ``converge_by`` block does not deal with idempotency and will set the ``update_by_last_action`` to ``true`` everytime the block executes.  By wrapping this block with an idempotency check like ``converge_if_changed`` or ``unless File.exist?("/tmp/foo")`` this will stop ``update_by_last_action`` from being set.
+The ``converge_by`` block is also responsible for setting ``update_by_last_action``, however it does not deal with idempotency and will set the ``update_by_last_action`` to ``true`` everytime the block runs.  By wrapping this block with an idempotency check like ``converge_if_changed`` or ``unless File.exist?("/tmp/foo")`` this will stop ``update_by_last_action`` from being set.
```
EasyAsABC123: @lamont-granquist not dropped moved down here

chef_master/source/custom_resources_notes.rst
EasyAsABC123: @lamont-granquist commented that it was moved down to the next paragraph, as i felt it made more sense in context with idempotency

chef_master/source/custom_resources_notes.rst
```diff
@@ -181,9 +181,9 @@ If you do need to write code which mutates the system through pure-Ruby then you
      end
    end
 
-The ``converge_by`` block gets why-run correct and will just touch "/tmp/foo" instead of actually doing it. The ``converge_by`` block is also responsible for setting ``update_by_last_action``.
+When the ``converge_by`` block is run in why-run mode, it will only log ``touch "/tmp/foo"`` and will not run the code inside the block. 
 
-In order to use ``converge_by`` correctly you must ensure that you wrap the ``converge_by`` with an idempotency check otherwise your resource will be updated every time it is used and will always fire notifications on every run.
+The ``converge_by`` block is also responsible for setting ``update_by_last_action``, however it does not deal with idempotency and will set the ``update_by_last_action`` to ``true`` everytime the block runs.  By wrapping this block with an idempotency check like ``converge_if_changed`` or ``unless File.exist?("/tmp/foo")`` this will stop ``update_by_last_action`` from being set.
```
lamont-granquist: This is a run-on sentence, and you've also dropped the fact that it fires notifications on every run when you forget it.   Also `converge_if_changed` wraps a `converge_by` block as well, so those are mutually exclusive, you shouldn't use both.   This should probably read more like:


    The ``converge_by`` block does not do any checking for idempotency and always sets ``updated_by_last_action``.  A
    ``converge_by`` block that is not wrapped in an idempotency check will always cause the resource to be updated, and
    will always cause notifications to fire.  A properly written resource should wrap all ``converge_by`` checks with an
    idempotency check, or the ``converge_if_changed`` API should be used instead.   The ``converge_if_changed`` API
    wraps a ``converge_by`` block with an idempotency check for you.

chef_master/source/custom_resources_notes.rst
lamont-granquist: yes, actually i see that now...

chef_master/source/custom_resources_notes.rst
EasyAsABC123: @lamont-granquist i didn't know `converge_if_changed` wrapped `converge_by` that should be in the other docs.  Awesome!

chef_master/source/custom_resources_notes.rst
EasyAsABC123: @lamont-granquist I changed it to API but i feel like it should be `block`

chef_master/source/custom_resources_notes.rst
lamont-granquist: some awkward code reading:

https://github.com/chef/chef/blob/master/lib/chef/provider.rb#L269,L285

all the code paths there wind up with the block getting passed into `converge_by`

EasyAsABC123: @lamont-granquist thanks for that! i was using these all wrong, greatly appreciated.

chef_master/source/custom_resources_notes.rst
```diff
@@ -181,9 +181,13 @@ If you do need to write code which mutates the system through pure-Ruby then you
...
+When the ``converge_by`` block is run in why-run mode, it will only log ``touch "/tmp/foo"`` and will not run the code inside the block. 
 
-In order to use ``converge_by`` correctly you must ensure that you wrap the ``converge_by`` with an idempotency check otherwise your resource will be updated every time it is used and will always fire notifications on every run.
+The ``converge_by`` block does not do any checking for idempotency and always sets ``updated_by_last_action``.  A
+``converge_by`` block that is not wrapped in an idempotency check will always cause the resource to be updated, and
+will always cause notifications to fire.  A properly written resource should wrap all ``converge_by`` checks with an
+idempotency check, or the [``converge_if_changed``](https://github.com/chef/chef-web-docs/blob/master/chef_master/source/custom_resources.rst#converge_if_changed) block should be used instead.   As the ``converge_if_changed`` API
+wraps a ``converge_by`` block with an idempotency check for you.
```
lamont-granquist: There's a dangling sentence fragment here, and I think there needs to be a connection between the negative effects (notifications always firing), to the solution (wrap in idempotency), and converge_if_changed then probably needs to stand alone as its own sentence.  Something like:

```
The ``converge_by`` block does not do any checking for idempotency and always sets ``updated_by_last_action``. 
A ``converge_by`` block that is not wrapped in an idempotency check will always cause the resource to be updated,
and will always cause notifications to fire.  To prevent this, a properly written resource should wrap all
``converge_by`` checks with an  idempotency check.  The
[``converge_if_changed``](https://github.com/chef/chef-web-docs/blob/master/chef_master/source/custom_resources.rst#converge_if_changed)
block may be used instead  which will wrap a ``converge_by`` block with an idempotency check for you.
```

lamont-granquist:
this is basically good to go technically :+1:

lamont-granquist (approved):

chef_master/source/custom_resources_notes.rst
lamont-granquist: ^ that edit has some redundancy but is getting closer to it

kagarmoe:
tenor-31273509

chef_master/source/custom_resources_notes.rst
EasyAsABC123: @lamont-granquist it appears `converge_if_changed` will check all properties on my resource if i do the following:

```
# if X or Y changed i'll do this block
converge_if_changed do
  # do one thing with X property
  # do one thing with Y property
end
```

PR #708, opened 2017-06-06
Remove duplicate from_email param in docs

tescalada:
The param from_email was defined twice, with different values listed as the default. I kept the better description but updated the default to match the one I found in my /etc/opscode/chef-server-running.json.


davidwrede:
tenor-55833599

PR #714, opened 2017-06-08
dtag changes for solaris package replicated to examples guide

brewn:


brewn:
just a dtags replicate for #713

rhass (approved):

PR #718, opened 2017-06-13
Adds example user data (post deployment) scripts in bash and powershell

grdnrio:
This PR adds examples of user data that can be run as part of an infrastructure deployment (AWS etc) to automate the bootstrapping and first run of a node.


jjasghar:
Great addition, we should get this is ASAP. 👍

simfish85 (commented):
👍 great addition

brewn:
tenor-235227792

PR #723, opened 2017-06-14
make wording more explicit for subscribes actions

brewn:


thommay:
Not sure this is quite right. We need to make clear that given:

```
file "/etc/foo"

service "foo" do
  subscribes :reload, "file[/etc/foo]"
end
```

what happens is that the `foo` service gets reloaded whenever the `/etc/foo` file is changed by chef. We don't affect any other resource, and we certainly don't affect the `/etc/foo` file resource.

timrchavez:
@thommay One confusing aspect of `subscribes` is that you are effectively subscribing to _any_ state change on a resource which is very subtle. I think people (maybe just new to chef) may intuit from `:action, :resource` that you are subscribing to a specific `:action` on `:resource` due to how it reads. In your example the `:reload` action does not make sense in the context of a file, so it illustrates the interface as its intended to be used, but what if it were two services where one can only reload after another reloads? Then it becomes a bit more ambiguous. As an aside, it seems like it would be nice to subscribe to an explicit `:action` because that seems more symmetrical with notifying an explicit `:action`

lamont-granquist:
yeah but notifications always fire if the resource us updated by any action.  we don't ever distinguish which action fires on the source resource and the action always refers to the target resource.  but when we do a subscribes we move the statement to the target resource.

and you'd need to come up with a solid use case for a *job* that you want to accomplish with the extension to the API.   changing the shape of the API based on theoretical concerns is likely to make it more complicated, whereas the existing rules are simple (once you understand that the action *always* applies to the target).  since nobody ever wants this functionality with notifies, i will bet that nobody wants this feature with subscribes either.  if you 'fix' one, you should 'fix' the other, and i've never seen that bug report for notifies, and notifies is used a lot more.

and if you really, really needed it, you could introduce two differently named resources to manage the same logical resource, but with different actions and then only subscribe to one of them.

thommay (approved):

PR #726, opened 2017-06-15
Fixing the Supermarket API documentation.

iennae:
There is a subtle difference between the sets of controllers available on the Supermarket
between https://supermarket.chef.io/cookbooks and https://supermarket.chef.io/api/v1.

https://supermarket.chef.io/cookbooks, /search, /tools are the html interface that users interact directly with via the website. The API that folks can curl against are what is behind /api/v1/INTERFACENAME.

If someone were to use the /cookbooks interface versus /api/v1/cookbooks for example they will have a different experience. One example is that /cookbooks doesn't have a user query term but /api/v1/cookbooks does.

Reference: [https://github.com/chef/supermarket/tree/master/src/supermarket/app/controllers](https://github.com/chef/supermarket/tree/master/src/supermarket/app/controllers)

Before merging, @robbkidd should give a 👍 .

/universe is the one intentional API endpoint not modified as it is available on /universe and returns a JSON response as expected.

### DIFF
--- a/chef_master/source/supermarket_api.rst
+++ b/chef_master/source/supermarket_api.rst
@@ -40,7 +40,7 @@ This method has no parameters.
 
 .. code-block:: none
 
-   POST /cookbooks/COOKBOOK_NAME
+   POST /api/v1/cookbooks/COOKBOOK_NAME
 
 **Response**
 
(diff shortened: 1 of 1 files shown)


chef_master/source/supermarket_api.rst
```diff
@@ -807,5 +806,10 @@ The response will return details for a user, including their name, Chef username
          "bacon": "https://supermarket.chef.io/api/v1/cookbooks/bacon"
          "chef-sugar": "https://supermarket.chef.io/api/v1/cookbooks/chef-sugar"
        }
-     }
+     }, 
+     "tools": {
+       "owns": {
+         "bacon_tool": "https://supermarket.chef.io/api/v1/tools/bacon_tool"
```
robbkidd, line 812: Mmmm. 🥓

robbkidd (approved):
I laughed. I cried. It was the "feel good" PR of the summer.

:shipit:

brewn:
tenor-124157559

PR #727, opened 2017-06-15
Update Sphinx to 1.6.2

davidwrede:
This is a LONG overdue update to the most recent version of Sphinx.

Later versions of Sphinx throw warnings if the pygment lexer cannot effectively parse code blocks. Because we treat all warnings as errors during build time, any "broken" code blocks would have to be fixed before the docs could be successfully built. 

For educational purposes, sometimes it's important to show non-working code blocks that contain pseudo code, even if they aren't 100% syntactically correct. In that vein, this PR removes some highlighting from a few of these code blocks in order for the new version of Sphinx to not throw warnings.

Newer versions of Sphinx also changed how they handled template overrides (either that, or we simply took advantage of a bug in earlier versions). So this PR sets the override correctly to use our custom search page instead instead of the basic search page.


brewn:
tenor-115791160

PR #728, opened 2017-06-15
Secrets loaded with IO.read need newline characters stripped

taharah:
Encrypted data bag secrets can contain newline characters that need to be removed in order to successfully load the data bag item.

### DIFF
--- a/chef_master/source/data_bags.rst
+++ b/chef_master/source/data_bags.rst
@@ -549,7 +549,7 @@ To load the secret from a file:
 
 .. code-block:: ruby
 
-   data_bag_item('bag', 'item', IO.read('secret_file'))
+   data_bag_item('bag', 'item', IO.read('secret_file').strip)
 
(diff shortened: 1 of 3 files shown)


brewn:
Thanks @taharah, I've PR'd some changes here:

https://github.com/taharah/chef-web-docs/pull/1

brewn:
Err, looks like this may already be covered here:

https://github.com/chef/chef-web-docs/pull/729

taharah:
Using the method `Chef::EncryptedDataBagItem.load_secret` clashes with the Foodcritic rule [FC086](http://www.foodcritic.io/#FC086). The `load_secret` method itself performs a strip when reading the file ([encrypted_data_bag_item.rb#L147](https://github.com/chef/chef/blob/master/lib/chef/encrypted_data_bag_item.rb#L147)), so this should achieve the same functionality while not violating FC086.

PR #729, opened 2017-06-16
Have data_bag_item use Chef::EncryptedDataBagItem.load_secret instead…

itmustbejj:
… of IO.read

A small change to use the load_secret method, which supports both IO.Read, and well as Kernel.open for remote key support.

Signed-off-by: Josh Hudson <jhudson@chef.io>

### DIFF
--- a/chef_master/source/data_bags.rst
+++ b/chef_master/source/data_bags.rst
@@ -549,7 +549,7 @@ To load the secret from a file:
 
 .. code-block:: ruby
 
-   data_bag_item('bag', 'item', IO.read('secret_file'))
+   data_bag_item('bag', 'item', Chef::EncryptedDataBagItem.load_secret(secret_file))
 
 To load a single data bag item named ``admins``:
 
--- a/chef_master/source/release_notes.rst
+++ b/chef_master/source/release_notes.rst
@@ -7507,7 +7507,7 @@ To load the secret from a file:
 
 .. code-block:: ruby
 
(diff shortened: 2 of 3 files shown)


spion06:
+1

brewn:
This method works, but the preferred style is http://www.foodcritic.io/#FC086 for readability.

spion06:
@brewn FC086 gives no recommendations on loading a secret from a path. Only on loading the data bag itself. Are you suggesting `IO.read` is the appropriate way to load a secret?

itmustbejj:
It was my impression that the FC086 violation was a false positive. The spirit of FC086, as I understand it, is to suggest relying on the `data_bag_item` abstraction, rather than calling `Chef::EncryptedDataBagItem.load` or `Chef::DataBagItem.load`. I am doing this in the example.

You can see the check is doing an xpath search for "EncryptedDataBagItem" or "DataBagItem", causing a false match with how I'm loading the secret.
https://github.com/Foodcritic/foodcritic/blob/master/lib/foodcritic/rules/fc086.rb#L4

tas50:
This works technically, but it's not required for most uses. data_bag_item is a pretty fancy helper and it automatically detects encrypted data bags. When it finds one it loads the secret file from the default location and decrypts it automatically. The only reason you would need to pass in the secret is if you used non-default locations.

https://github.com/chef/chef/blob/db57131ad383076391b9df32d5e9989cfb312d58/lib/chef/dsl/data_query.rb#L58

PR #734, opened 2017-06-19
Manage: document reCAPTCHA settings

srenatus:
These will be part of the next Manage release. Since it'll happen _real soon_, there's no need to hold this back.


davidwrede:
tenor-73672358

srenatus:
@davidwrede Thanks for merging. I've just realised that I've forgotten to add any indication of "minimal version required" there 😖 *But* since this is Manage, I think we can let that slip. Ok?

davidwrede:
Probably okay in this instance. Thanks for checking @srenatus.

PR #736, opened 2017-06-20
Document new chef-server-ctl set-secret option.

phiggins:
Documentation for the new option added to `chef-server-ctl set-secret` in chef/chef-server#1313.

### DIFF
--- a/chef_master/source/ctl_chef_server.rst
+++ b/chef_master/source/ctl_chef_server.rst
@@ -680,6 +680,17 @@ There are various ways to pass the secret to this command:
       Enter ldap bind_password:    (no terminal output)
       Re-enter ldap bind_password: (no terminal output)
 
+**Options**
+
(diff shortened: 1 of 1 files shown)


blakestier:
👍 to ship after #1313

brewn:
For reference: https://github.com/chef/chef-server/pull/1313

Thanks for the update!

srenatus (approved):
The additions looks good -- just a note:
The option will exist in the _next version_ of Chef-Server, so, it's only available for >= 12.15.9. This isn't scheduled too soon, as far as I can tell (anything could happen, though), do we need some "added in version xyz" indication?

brewn:
@srenatus Yep that's a good idea - I'll add a note to this before it gets merged.

afiune (approved):

afiune:
Linking the PR that will add this functionality: https://github.com/chef/chef-server/pull/1313

PR #742, opened 2017-06-22
Remove FIPS note that referred to an old release

btm:
ChefDK 1.3.40 had a FIPS bug on Windows which was fixed in omnibus-software#823
and released ten days later in ChefDK 1.3.43. This note is cruft and should
have been removed then.


kagarmoe:
image

PR #744, opened 2017-06-23
Work around a bug in data collector token setting

sean-horn:
The Chef Server 12.15.8 data collector token must be set directly with the `chef-server-ctl set-secret` command in order to be changed.


brewn:
Thanks @sean-horn - some style adjustments here:

https://github.com/chef/chef-web-docs/pull/745

I tried to change the title to make the issue easily identifiable.

PR #758, opened 2017-06-29
Minor grammatical correction for execute resource

wheatevo:
This is just a tiny change to make the first sentence in the Syntax section more grammatically correct. Please let me know if there is a separate standard that is followed that would make `A resource` more desirable/correct.


brewn:
Thanks @wheatevo !

Issue #397, opened 2017-01-13
Add definitions for knife node status values


davidwrede:
The values are listed, but no definitions or additional context are provided.

PR #763, opened 2017-06-30
knife node status definitions

brewn:


brewn:
#397

iennae:
https://docs.chef.io/api_push_jobs.html#node-states needs to be in sync with information about the states I think as well.  The content from this PR looks like it matches the documentation from https://github.com/chef/opscode-pushy-server/blob/master/docs/rest-api.org. 

(knife node status is exercising that API)

brewn:
@iennae That's a good idea, but I'm going to create a separate issue / PR for that, as this one is already wandering past its scope with the inclusion of jobs status stuff.

PR #776, opened 2017-07-06
Example code for group_dn was mismatched to text above it

jblaine:
Pretty sure this is correct for the intention. It's certainly not any worse than what was there.


brewn:
Thanks @jblaine! I've submitted a PR that replicates this update across the tagged regions of our documentation:

https://github.com/jblaine/chef-web-docs/pull/1

Once this is merged, your contribution should pass the Travis check and we can get this updated.

PR #778, opened 2017-07-06
Change --chef-dk-version to --chefdk-version fix typo

jpstecroix:
when you install automate 1.5.46 and you want to add a runner node with a specific version the right option is --chefdk-version


brewn:
@jpstecroix nice catch! Thanks for your help.

PR #780, opened 2017-07-07
Automate backups should be initiated outside root-controlled directories

sean-horn:
root-only working directories cause failures during `automate-ctl create-backup` runs as the chpst command immediately tries to change into the user chef-pgsql's home directory and fails, causing follow on problems with the backup.

Zendesk 15110


davidwrede:
tenor-129673764

PR #785, opened 2017-07-10
add closing bracket to subscribes example code

mburns:
It appears like the example snippet for `subscribes` has been used in a bunch of places but is missing the closing `]`.


brewn:
@mburns Nice catch! Thanks for the fix.

Issue #629, opened 2017-04-28
missing documentation on signing protocol version 1.3


juliandunn:
This page https://docs.chef.io/auth.html#authentication-headers has some outdated information about the client-server signing protocol. In particular, it does not reflect that there is a version 1.3 of the signing protocol available which uses SHA256 instead of SHA1 as the hash algorithm, and how to set this.

To wit, in `client.rb`: 

`authentication_protocol_version = '1.3'`

Or for `knife.rb`:

`knife[:authentication_protocol_version] = '1.3'`

There are probably other places in our documentation (e.g. `knife.rb` settings file) that also need an update.

juliandunn:
Note that not all versions of the Chef Server support protocol version 1.3. When you get to this ticket, we'll need to a little spelunking to figure out what versions this started being available in.

PR #790, opened 2017-07-11
Adds steps for enabling TLS 1.3

brewn:


brewn:
#629

juliandunn (changes requested):
Authentication protocol 1.3 is *not* a TLS version; it refers to the custom (Chef-developed) protocol that clients and servers use to sign their requests.

chef_master/source/config_rb.rst
```diff
@@ -73,6 +80,11 @@ This configuration file has the following settings:
 
       data_bag_encrypt_version 2
 
+``fips``
+  Allows OpenSSL to enforce FIPS-validated security during the chef-client run. Set to ``true`` to enable FIPS-validated security.
+
+  Changed in Chef server 12.13 to expose FIPS runtime flag on RHEL.
```
juliandunn: I don't know what that means

brewn: This was pulled in as part of a replication between the settings in the knife.rb documentation and those in config.rb, as the two were beginning to splinter, with knife.rb getting more updates. 

The "Changed in Chef server 12.13 to expose FIPS runtime flag on RHEL." line does seem like a non-sequitur here, though. I think it's meant to be a very roundabout way of saying that FIPS mode is only supported on RHEL for Chef server.

While the [notes](https://github.com/chef/chef-server/commit/4160a18946b07b9559786abfb32a1976f52008ca) say the same thing, I suspect that if FIPS is supported for server on RHEL it likely works on CentOS and Oracle as well..

**Edit:** Confirmed platform support with @tyler-ball

PR #792, opened 2017-07-12
Format subtitle in audit_supported_configurations.rst

algaut:
'Fetch From Automate via Chef Server' subtitle was not correctly formatted


brewn:
@algaut Thanks for catching that!

PR #812, opened 2017-07-21
powershell_script examples incorrectly refer to batch in likely copy-paste error

btm:


brewn:
@btm Looks like this is pulled in from an include in `resource_batch.rst`. I'll commit a change to split the two.

brewn:
Thanks for catching this! Have a good weekend. 
tenor-180136912

PR #813, opened 2017-07-21
Rework supported platforms

tas50:
Add FreeBSD 11 as as a secondary platform as we have a package for it, but we're not entirely testing it yet. Remove Debian 6 which is EOL, but add 9 which we're testing in the cookbook ecosystem now.

We claim that secondary platforms have packages, but we don't test them. Not sure why Arch in Gentoo got in here then since they have no packages.


brewn:
@tas50 I don't see a package for FreeBSD 11 on https://downloads.chef.io/chef/13.2.20#freebsd

Also re: Gentoo and Arch, I think OmniOS falls into the same category. This doc has a major overhaul coming soon, and I'll make sure this stuff gets incorporated.

tas50:
For FreeBSD we have a compatible package in that the FreeBSD 9/10 packages work fine on 11. That's the same for a lot of platforms such as Scientific or Oracle where we don't have a specific package, but the generic RHEL packages work. The same thing happens with OmniOS where Solaris stuff works I believe. Gentoo and Arch are going to require the user to install from Rubygems which we highly discourage

brewn:
We're holding off on any changes to this guide while it gets an overhaul. I'm tagging this so these changes can be incorporated.

PR #819, opened 2017-07-24
(PR818) adds "elevated" property to powershell_script

brewn:


brewn:
Closing in favor of #829 #831

Issue #78, opened 2016-06-29
Incorrect class name `Chef::Provider::Service::Init::Debian`


jhulten:
Unless I am misreading the source, there is no `Chef::Provider::Service::Init::Debian`, only `Chef::Provider::Service::Debian`. Am I incorrect?

lamont-granquist:
believe that is correct

PR #821, opened 2017-07-25
Updates to resource provider section

brewn:


brewn:
#78

chef_master/source/resource_apt_package.rst
```diff
@@ -302,11 +302,30 @@ The chef-client will determine the correct provider based on configuration data
...
+     pkg_resource = case node['platform_family']
+       when "debian"
+         :dpkg_package
+       when "fedora", "rhel", "amazon"
+         :rpm_package
+       end
+
+     pkg_path = ( pkg_resource == :dpkg_package ) ? "/tmp/foo.deb" : "/tmp/foo.rpm"
+
+     build_resource(pkg_resource, pkg_path) do
```
lamont-granquist: should be `declare_resource`.

`build_resource` doesn't put it into the resource collection.  you'd use that if you wanted to build a resource, then later explicitly call an action on with `run_action(:whatever)` which should not be a common use case worth documenting.  generally you want to stick to encouraging people to always call `declare_resource` (in fact `build_resource` may be one of those APIs that really shouldn't be put in front of customers since it'll confuse them and be wrong in 99% of their cases).

lamont-granquist:
yeah just `s/build_resource/declare_resource/g` and ship it.

lamont-granquist (approved):

PR #824, opened 2017-07-27
Remove a few node.set / node.normal usages

tas50:
This will work on Chef 14. The old examples would not


tas50:
Tagging @robbkidd since I'm touching Supermarket stuff

robbkidd:
Because I'm behind the times, I had to go read https://docs.chef.io/deprecations_attributes.html#set-and-set-unless to know about the deprecation.

robbkidd (approved):
👍

PR #825, opened 2017-07-28
Misc fixes to resources

tas50:


brewn:
@tas50 This required a pretty crazy amount of tag fixing..I think everything ended up in the right place, but you might one to take a quick look at the finished product before I merge it.

chef_master/source/resource_yum.rst
```diff
@@ -472,9 +472,9 @@ or:
 
 .. tag resource_package_handle_cookbook_file_and_package
 
-.. To handle cookbook_file and yum_package when both called in the same recipe
```
tas50: This one we don't actually want to replicate. This actually should be yum_package since it's in the yum resource page

brewn: I'll switch it back to the old tag.

brewn: This is going to require some more shuffling around, a lot of these tagged regions are referring to yum_package instead of package.

PR #833, opened 2017-08-01
Fix typo in tagging example

neilmb:
This fixes a typo that used an apostrophe inside a single-quoted string.


brewn:
Thanks @neilmb! tenor-34416232

PR #837, opened 2017-08-02
bolded product titles and minor copy changes

brewn:
I've been looking at this doc a lot recently, and it's always visually disorienting when the titles/product names are the same font as the descriptions.


davidwrede:
tenor-107259470

PR #839, opened 2017-08-02
(DO NOT MERGE) Updates Makefile to only build changed files (after first time)

brewn:
The way this works:

- An initial `make` will still generate the full doc set under `chef_master/source`. The state of the files is hashed in `build/.buildinfo`
- Subsequent `make`s within the same branch will only regenerate the updated files.


brewn:
This second commit creates a makefile under `doctools` for [sphinx autobuild](https://pypi.python.org/pypi/sphinx-autobuild) that can be used with `make -f`. It does live reload and spins up a temporary web server on port 8000.

brewn:
Closing this because it's going to be a moot point.

PR #840, opened 2017-08-03
Corrected Wow6464Node reference to Wow6432Node

Sam-Martin:


brewn:
Thanks @Sam-Martin!

tenor-155966417

PR #842, opened 2017-08-03
Edit chefdk version flag to show correct syntax

bnwoods:
This doc page shows the incorrect flag for --chefdk-version in the install-runners command. It shows the incorrect flag --chef-dk-version.


brewn:
Good catch, @bnwoods! Thanks. 
tenor-53604846

PR #855, opened 2017-08-10
Kg/kibana

kagarmoe:


davidwrede:
Note (with updated text) already covered in #848, so I'm closing out this PR.

PR #864, opened 2017-08-11
Added VMware page

jjasghar:
- Added a overview of VMware and Chef integrations for docs.chef.io

### DIFF
--- a/chef_master/source/vmware.rst
+++ b/chef_master/source/vmware.rst
@@ -1,0 +1,527 @@
+=====================================================
+Chef and VMware
+=====================================================
+
+`[edit on GitHub] <https://github.com/chef/chef-web-docs/blob/master/chef_master/source/vmware.rst>`__
+
+.. note:: This page collects information about Chef that is specific to using Chef with VMware.
+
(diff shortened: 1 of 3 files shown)


chef_master/source/vmware.rst
```diff
@@ -0,0 +1,527 @@
...
+Here are some basic usage examples to help get you started.
+
+This clones from a VMware template and bootstraps chef into it. It uses the generic DHCP options.
+
+.. code-block:: bash
+
+   $ knife vsphere vm clone MACHINENAME --template TEMPLATENAME --bootstrap --cips dhcp
+
+This clones a vm from a VMware template bootstraps chef, then uses a Customization template
+called "SPEC" to help bootstrap. Also calls a different SSH user and Password.
```
brewn: If the template is called `SPEC`, is that what should be listed under `TEMPLATENAME`? Or are the two entirely separate?

brewn: Nevermind, I was reading this wrong, thought it was referring to the example above it, instead of the one below.

jjasghar:
🤘 awesome. thanks for the updates.

PR #867, opened 2017-08-14
Missing single quote in hash key

mrvovanness:


brewn:
Good catch! Thanks @mrvovanness

PR #879, opened 2017-08-16
replace "mv" with canonical way to install new repos

juliandunn:


brewn:
I don't have a RHEL install to test, but at least on CentOS 7 `yum-config-manager` requires the `yum-utils` package, which isn't installed by default.

juliandunn:
huh... it's on the AWS CentOS 7 image as well as the `centos:7` Docker image by default.

PR #891, opened 2017-08-22
(incomplete) airgap guide

brewn:


chef_master/source/install_chef_air_gap.rst
```diff
@@ -0,0 +1,602 @@
...
+
+   cd /tmp/
+   wget http://packages.example.com/chef_13.2.20-1_amd64.deb
+   dpkg -i chef_13.2.20-1_amd64.deb
+
+The install script should be accessible from your artifact store.
+
+Chef server
+=====================================================
+In this section you'll create install the Chef server, and create your organization and user.  Note that in order to configure Supermarket later in this guide, you will need a user that is a member of the ``admins`` group.
```
galenemery: In this section you'll create install...

We don't need "create"

chef_master/source/install_chef_air_gap.rst
```diff
@@ -0,0 +1,602 @@
...
+
+.. note:: If you intend to use Chef Automate, create the ``delivery`` user and add it to your organization during this step.
+
+#. Download the package from https://downloads.chef.io/chef-server/.
+#. Upload the package to the machine that will run the Chef server, and then record its location on the file system. The rest of these steps assume this location is in the ``/tmp`` directory.
+
+#. .. tag install_chef_server_install_package
```
galenemery, line 72: Does this component include the info on manually installing add-ons (like chef-manage?)  If not, we should ensure its included in this guide.

chef_master/source/install_chef_air_gap.rst
```diff
@@ -0,0 +1,602 @@
...
+* ``-x`` defines the (ssh) user name: ``ubuntu``
+* ``--sudo`` ensures that sudo is used while running commands on the node during the bootstrap operation
+
+When the bootstrap operation is finished, do the following:
+
+#. Edit the node to add the wrapper cookbook's ``/recipes/default.rb`` recipe to the run-list:
+
+   .. code-block:: bash
+
+      $ knife node edit supermarket-node
```
galenemery: This has issues when run from a windows workstation.  I suggest `knife node run_list set supermarket-node recipe[my_supermarket_wrapper::default]`

chef_master/source/install_chef_air_gap.rst
```diff
@@ -0,0 +1,602 @@
...
+   .. code-block:: bash
+
+      $ sudo chef-client
+
+Connect to Supermarket
+-----------------------------------------------------
+To reach the newly spun up private Chef Supermarket, the hostname must be resolvable from a workstation. For production use, the hostname should have a DNS entry in an appropriate domain that is trusted by each user's workstation.
+
+#. Visit the Chef Supermarket hostname in the browser. A private Chef Supermarket will generate and use a self-signed certificate, if a certificate is not supplied as part of the installation process (via the wrapper cookbook).
+#. If an SSL notice is shown while connecting to Chef Supermarket via a web browser, accept the SSL certificate. A trusted SSL certificate should be used for  private Chef Supermarket that is used in production.
```
galenemery: Assuming you're using a self-signed certificate.

chef_master/source/install_chef_air_gap.rst
```diff
@@ -0,0 +1,602 @@
...
+
+   data_collector['root_url'] = 'https://my-automate-server.mycompany.com/data-collector/v0/'
+   data_collector['token'] = 'TOKEN'
+
+where ``my-automate-server.mycompany.com`` is the fully-qualified domain name of your Chef Automate server, and
+``TOKEN`` is either the default value or the token value you configured in the `prior section <#configure-a-data-collector-token-in-chef-automate>`__.
+
+Save the file and run ``chef-server-ctl reconfigure`` to complete the process.
+
+Additional configuration options include:
```
galenemery, line 594: We may want to mention `opscode_erchef['max_request_size']` as a tunable.  Its useful when proxying audit runs of large inspec profiles via the chef-server.

galenemery (commented):
This is amaze balls.  A few clarifying things in there based on my experience, but otherwise a very solid guide.

chef_master/source/install_chef_air_gap.rst
brewn: Because this guide assumes someone might be reading it via pdf within an air-gapped network, it pulls in a lot of install information from other guides. As a result, it's quite long for a procedural doc - and I'm wary of making it any longer. 

We're going to have to get a sense of what add-ons and features are most in demand for users in an air-gapped environment (assuming that the offline aspect affects the setup process somehow) and then potentially create a subset of documentation for air-gapped networks.

chef_master/source/install_chef_air_gap.rst
brewn: Thanks for all your help with this, @galenemery!

PR #904, opened 2017-08-31
Break up knife getting started / configuration better

tas50:
The knife documentation has a bit of overlap right now that makes it hard to get the complete picture without reading through several docs. There's a knife doc that includes a brief overview and all the subcommands, there's a getting started doc that tells you how to setup part of your config and then explains how knife sub commands work, then there's the knife.rb file which has some of the same stuff again.  This breaks up documentation into the following:

Knife: Overview of what knife is, how the commands are structured, what the subcommands are, and an overview of plugins
Knife Setup: Information on configuring knife including what the knife.rb file is and where we load it from (wasn't documented before)
Knife.rb: A breakdown of each of the knife.rb config options


tas50:
It should be noted that this isn't the final state of these docs, but I think this gets it much better and it's something we could merge right now and get value from.

chef_master/source/knife_setup.rst
```diff
@@ -0,0 +1,131 @@
...
+Setting up Knife
+=====================================================
+`[edit on GitHub] <https://github.com/chef/chef-web-docs/blob/master/chef_master/source/knife_using.rst>`__
+
+The knife command line tool must be configured to communicate with the Chef Server as well as any other infrastructure within your organization. This is done initially during the workstation setup, but subsequent modifications can be made using the knife.rb configuration file.
+
+Knife.rb Configuration File
+=====================================================
+
+Knife is configured using a knife.rb configuration, which contains configuration for both the knife command line tool as well as any installed knife plugins. See :doc:`knife.rb </config_rb_knife>` for a complete list of configuration options in the knife.rb file.
```
brewn, line 30: This is linking to the `config_rb_knife` file, but that file is redirected back to `knife_setup` ..is the intention to get rid of `config_rb_knife` altogether?

tas50: Let me check on that.

tas50: I just built locally and this links to the doc with all the common config items. That should be good

chef_master/source/knife.rst
```diff
@@ -16,7 +16,7 @@ knife is a command-line tool that provides an interface between a local chef-rep
 
 .. end_tag
 
-.. note:: The Knife Quick Reference provides an all-in-one quick reference of knife commands. View a web-based PNG file here: |url docs_knife_png|. Or download the source files from here: |url docs_repo_qr|. Print the front/back source files and laminate them for best effect.
```
tas50, line 19: You don't randomly laminate tech guides? ;)

PR #905, opened 2017-08-31
Remove references to knife cookbook test

tas50:
We don't suggest or support this anymore. It doesn't entirely work and there's better tooling out there to perform syntax checking at this point (cookstyle/chefspec)


iennae:
Since the option still exists even though it has a warning "DEPRECATED" I think in the documentation it should have a deprecated note prior to just being deleted.

brewn:
Agreed that this feature should be documented somewhere if it still exists, so I've added deprecation warnings. Still removing it from the client upgrade process list.

PR #906, opened 2017-08-31
Remove knife config for chef container

tas50:
The chef container functionality is EOL beyond EOL. There's no need for this config option at this point.


davidwrede:
tenor-65012316

PR #907, opened 2017-08-31
Remove ruby requirement from chef-client

tas50:
We're constantly changing the version of ruby required for gem installs, but we don't support gem installs. We only support omnibus installs and for those this just doesn't matter to the end user. We package it all up and they just run the installer.


davidwrede:
tenor-145908702

davidwrede:
Change looks good

PR #914, opened 2017-09-06
Fix incorrect source example for chef server

devosnw:
I believe the example for getting Berkshelf to work with open source chef is incorrect. I think the syntax should be similar to the local Chef repository example below. If I am wrong let me know and I will withdraw this PR but this is how I had to configure my `Berksfile` on my system to get it working.

### DIFF
--- a/chef_master/source/berkshelf.rst
+++ b/chef_master/source/berkshelf.rst
@@ -96,7 +96,7 @@ To add a Chef Server:
 .. code-block:: ruby
 
    source "https://supermarket.chef.io"
-   source :chef_server
+   source chef_server: "https://chef-server.example.com"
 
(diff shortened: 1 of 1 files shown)


brewn:
Hi @devosnw, 

The `:chef_server` source will pull from the information in your `knife.rb` file. Have you confirmed that the Chef server config in that file is correct? `knife ssl check` is an easy way to be sure. 

Also, what version of ChefDK are you using?

devosnw:
Ok got it. Thanks for clarifying that. Maybe that sentence below about using the authentication details in `knife.rb` should be moved up in the same section when talking about the chef server. I think that's where I got confused because I wasn't sure how berks was going to discover where the chef server is.

PR #915, opened 2017-09-06
[MSYS-439] Updated docs for alternate user support in remote_file resource

NimishaS:
w.r.t. https://github.com/chef/chef/pull/5832

### DIFF
--- a/chef_master/source/resource_remote_file.rst
+++ b/chef_master/source/resource_remote_file.rst
@@ -60,6 +60,9 @@ The full syntax for all of the properties that are available to the **remote_fil
      use_last_modified          TrueClass, FalseClass
      verify                     String, Block
      show_progress              TrueClass, FalseClass
+     remote_user                String
+     remote_password            String
+     remote_domain              String
      action                     Symbol # defaults to :create if not specified
    end
 
@@ -68,6 +71,7 @@ where
 * ``remote_file`` is the resource
 * ``name`` is the name of the resource block
 * ``action`` identifies the steps the chef-client will take to bring the node into the desired state
+* ``remote_user``, ``remote_password`` and ``remote_domain`` are used to authenticate to a remote server when accessing a file over SMB/CIFS specified as a UNC path. This is only supported on Windows systems.
 * ``atomic_update``, ``backup``, ``checksum``, ``force_unlink``, ``ftp_active_mode``, ``group``, ``headers``, ``inherits``, ``manage_symlink_source``, ``mode``, ``owner``, ``path``, ``provider``, ``rights``, ``source``, ``use_conditional_get``, ``use_etag``, ``use_last_modified``, ```show_progress`` and ``verify`` are properties of this resource, with the Ruby type shown. See "Properties" section below for more information about all of the properties that may be used with this resource.
 
 Actions
@@ -371,10 +375,25 @@ This resource has the following properties:
 
    Enable If-Modified-Since headers. Set to ``false`` to disable If-Modified-Since headers. To use this setting, ``use_conditional_get`` must also be set to ``true``. Default value: ``true``.
 
-``show_progess`` 
+``show_progess``
    **Ruby Types:** TrueClass, FalseClass
 
    Displays the progress of the file download. Set to ``true`` to enable this feature. Default value: ``false``.
+
+``remote_user``
+   **Ruby Types:** String
+
+   *Windows only:* The name of a user with access to the remote file specified by the `source` property. Default value: `nil`. The user name may optionally be specifed with a domain, i.e. `domain\user` or `user@my.dns.domain.com` via Universal Principal Name (UPN) format. The domain may also be set using the `remote_domain` property. Note that this property is ignored if `source` is not a UNC path. If this property is specified, the remote_password property is required.
+
+``remote_password``
+   **Ruby Types:** String
+
+   *Windows only:* The password of the user specified by the `remote_user` property. Default value: `nil`. This property is required if `remote_user` is specified and may only be specified if `remote_user` is specified. The `sensitive` property for this resource will automatically be set to `true` if `remote_password` is specified.
+
+``remote_domain``
+   **Ruby Types:** String
+
+   *Windows only:* The domain of the user specified by the `remote_user` property. Default value: `nil`. By default the resource will authenticate against the domain of the remote system, or as a local account if the remote system is not joined to a domain. If the remote system is not domain joined, it is necessary to authenticate as a local user on the remote system by setting the domain to ``.``, e.g. ``remote_domain "."``. The domain may also be specified as part of the remote_user property.
 
 ``verify``
    **Ruby Types:** String, Block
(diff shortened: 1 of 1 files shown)


NimishaS:
@btm , please review.

chef_master/source/resource_remote_file.rst
```diff
@@ -68,6 +71,7 @@ where
 * ``remote_file`` is the resource
 * ``name`` is the name of the resource block
 * ``action`` identifies the steps the chef-client will take to bring the node into the desired state
+* ``remote_user``, ``remote_password`` and ``remote_domain`` are used for accessing a remote UNC path on Windows. These properties are Windows specific only.
```
btm: * ``remote_user``, ``remote_password``, and ``remote_domain`` are used to authenticate to a remote server when accessing a file over SMB/CIFS specified as a UNC path. This is only supported on Windows systems.

chef_master/source/resource_remote_file.rst
```diff
@@ -371,11 +375,26 @@ This resource has the following properties:
...
-``show_progess`` 
+``show_progess``
    **Ruby Types:** TrueClass, FalseClass
 
    Displays the progress of the file download. Set to ``true`` to enable this feature. Default value: ``false``.
 
+``remote_user``
+   **Ruby Types:** String
+
+   *Windows only:* The user name of a user with access to the remote file specified by the `source` property. Default value: `nil`. The user name may optionally be specifed with a domain, i.e. `domain\user` or `user@my.dns.domain.com` via Universal Principal Name (UPN) format. It can also be specified without a domain simply as `user` if the domain is instead specified using the `remote_domain` attribute. Note that this property is ignored if `source` is not a UNC path. If this property is specified, the `remote_password` property **must** be specified.
```
btm: "The name of a user with access to the remote file"

"The domain may also be set using the `remote_domain` property."

"If this property is specified, the `remote_password` property is **required**."

chef_master/source/resource_remote_file.rst
```diff
@@ -371,11 +375,26 @@ This resource has the following properties:
...
+
+``remote_password``
+   **Ruby Types:** String
+
+   *Windows only:* The password of the user specified by the `remote_user` property. Default value: `nil`. This property is mandatory if `remote_user` is specified and may only be specified if `remote_user` is specified. The `sensitive` property for this resource will automatically be set to `true` if `remote_password` is specified.
```
btm: I'd probably say *required* instead of mandatory but no big deal.

chef_master/source/resource_remote_file.rst
```diff
@@ -371,11 +375,26 @@ This resource has the following properties:
...
+
+``remote_domain``
+   **Ruby Types:** String
+
+   *Windows only:* The domain of the user user specified by the `remote_user` property. Default value: `nil`. If not specified, the user and password properties specified by the `remote_user` and `remote_password` properties will be used to authenticate that user against the domain in which the system hosting the UNC path specified via `source` is joined, or if that system is not joined to a domain it will authenticate the user as a local account on that system. An alternative way to specify the domain is to leave this property unspecified and specify the domain as part of the `remote_user` property.
```
btm: I'd shorten this:

> If not specified, the user and password properties specified by the `remote_user` and `remote_password` properties will be used to authenticate that user against the domain in which the system hosting the UNC path specified via `source` is joined

to just:
By default the resource will authenticate against the domain of the remote system, or as a local account if the remote system is not joined to a domain.

And:
> An alternative way to specify the domain is to leave this property unspecified and specify the domain as part of the `remote_user` property.

to:

The domain may also be specified as part of the `remote_user` property.

chef_master/source/resource_remote_file.rst
```diff
@@ -600,6 +619,58 @@ To prevent the chef-client from re-downloading files that are already present on
 
 The desired approach just depends on the desired workflow. For example, if a node requires a new file every day, using the checksum approach would require that the local checksum be updated and/or verified every day as well, in order to ensure that the local checksum was the correct one. Using a conditional ``GET`` in this scenario will greatly simplify the management required to ensure files are being updated accurately.
 
+Access a remote UNC path on Windows
+-----------------------------------------------------
+The ``remote_file`` resource now supports the use of credentials on Windows when accessing a remote UNC path on Windows such as ``\\myserver\myshare\mydirectory\myfile.txt``. This
```
btm: This is documentation, not release notes, so drop the "now"

> The `remote_file` resource on Windows supports accessing files from a remote SMB/CIFS share. The file name should be specified in the `source` property as a UNC path.

chef_master/source/resource_remote_file.rst
```diff
@@ -600,6 +619,58 @@ To prevent the chef-client from re-downloading files that are already present on
...
+allows access to the file at that path location even if the Chef client process identity does not have permission to access the file. The new properties ``remote_user``, ``remote_domain``, and ``remote_password`` may be used to specify credentials with access to the remote file so that it may be read. Please check the details of the usage of these properties above.
```
btm: Credentials for authenticating to the remote system can be specified using the `remote_user`, `remote_domain`, and `remote_password` properties when the user that the Chef client is running does not have access to the remote file. More detail about the properties is provided in the Properties section above.

chef_master/source/resource_remote_file.rst
```diff
@@ -600,6 +619,58 @@ To prevent the chef-client from re-downloading files that are already present on
...
+
+**Note**: This feature is mainly used for accessing files between two nodes in different domains and having different user accounts.
```
btm: This is primarily for accessing remote files when the user that the Chef client is running as does not have sufficient access and alternate credentials need to be specified. If the user already has access, the credentials do not need to be specified.

chef_master/source/resource_remote_file.rst
```diff
@@ -600,6 +619,58 @@ To prevent the chef-client from re-downloading files that are already present on
...
+In case the two nodes are in same domain, ``remote_file`` resource does not need ``remote_user`` and ``remote_password`` specified because the user has the same access on both systems through the domain.
```
btm: In the case that the local system and remote system are in the same domain, the `remote_user` and `remote_password` properties often do not need to be specified as the user may already have access to the remote file share. However these properties may still be used if alternate credentials are needed to access that share.

chef_master/source/resource_remote_file.rst
```diff
@@ -600,6 +619,58 @@ To prevent the chef-client from re-downloading files that are already present on
...
+
+For example:
+
+Accessing file from a (different) domain account
```
btm: Accessing a file from a different domain account

chef_master/source/resource_remote_file.rst
```diff
@@ -600,6 +619,58 @@ To prevent the chef-client from re-downloading files that are already present on
...
+
+.. code-block:: ruby
+
+   remote_file "E://domain_test.txt"  do
```
btm: do we really need two forward slashes here (and below), or is `E:/domain_test.txt` sufficient? normally we have two backslashes because the backslash is an escape character for double-quoted strings, e.g. `E:\\domain_test.txt`

chef_master/source/resource_remote_file.rst
```diff
@@ -600,6 +619,58 @@ To prevent the chef-client from re-downloading files that are already present on
...
+     source  "\\\\myserver\\myshare\\mydirectory\\myfile.txt"
+     remote_user "domain\\username"
+     remote_password "password"
+   end
+
+Accessing file using a local account on the remote machine
```
btm: Accessing a file using a local account on the remote machine:

chef_master/source/resource_remote_file.rst
```diff
@@ -600,6 +619,58 @@ To prevent the chef-client from re-downloading files that are already present on
...
+
+.. code-block:: ruby
+
+   remote_file "E://domain_test.txt"  do
+     source  "\\\\myserver\\myshare\\mydirectory\\myfile.txt"
+     remote_domain "."
```
btm: we might want to add a note above that the `.` is required to specify the local user account. That could be missed here if this goes unexplained.

Like under the "remote_domain" property section something like "If the remote system is domain joined, but it is necessary to authenticate as a local user on the remote system, set the domain to `.`, e:g: ` remote_domain "."`.

btm (approved):

PR #918, opened 2017-09-07
Document support for multiple keys in apt_repository

gsreynolds:
From Chef client 13.4, multiple keys will be supported in apt_repository resource

Related: https://github.com/chef/chef/pull/6372


brewn:
Thanks for adding this, @gsreynolds

PR #920, opened 2017-09-07
Update ctl_chef_server.rst

richardnaik:
In newer versions of Chef Server, ha-status only returns an error message (`[ERROR] keepalived HA services not enabled, please configure keepalived according to documentation.`). As far as I can tell this has been the case since 12.9 when the old HA architecture was deprecated.

### DIFF
--- a/chef_master/source/ctl_chef_server.rst
+++ b/chef_master/source/ctl_chef_server.rst
@@ -142,6 +142,13 @@ This subcommand has the following syntax:
 
 ha-status
 =====================================================
+
+.. warning:: .. tag chef_license_note_current
+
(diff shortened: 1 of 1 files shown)


brewn:
@richardnaik This command is only meant to be used with an older DRBD HA setup. If you're using such a setup and you're running into this error, I'd encourage you to open a Chef server [issue](https://github.com/chef/chef-server/issues).

I do agree that the wording could be strengthened a bit to make clear that this command is DRBD / keepalive-specific, though. I'll add some clarification to it.

richardnaik:
Thanks.

PR #922, opened 2017-09-08
Adding xpack options being added to Automate 1.6.156

thomascate:
Adding docs for auth options when using ES X-Pack.


davidwrede:
tenor-65070506

PR #934, opened 2017-09-13
Renames to chef_client_overview and changes associated files

kagarmoe:


chef_master/source/chef_client_overview.rst
```diff
@@ -1,33 +1,16 @@
...
+Chef Client Overview
 =====================================================
 `[edit on GitHub] <https://github.com/chef/chef-web-docs/blob/master/chef_master/source/chef_client.rst>`__
 
-.. tag chef_client_summary
 
-A chef-client is an agent that runs locally on every node that is under management by Chef. When a chef-client is run, it will perform all of the steps that are required to bring the node into the expected state, including:
 
-* Registering and authenticating the node with the Chef server
-* Building the node object
```
brewn: I agree with removing the Ohai section, but reading through this I feel like it's necessary to mention that client runs Ohai at the start of every chef client run. Maybe add a bullet point with a link that says something like "Uses :doc:`Ohai </ohai>` to determine the system state."

PR #945, opened 2017-09-15
Update knife_role.rst

derekgroh:
syntax correction


brewn:
Thanks for the fix, @derekgroh!

PR #959, opened 2017-09-21
Specify correct ChefDK version

tpetchel:
If you specify `--chefdk-version 1.5`, you get this error:

```
Downloading the 1.5 version of the ChefDK. This make take a few minutes...
ERROR: You requested ChefDK version 1.5 via --chefdk-version but the package did not exist.
For a valid list of versions for your platform, see https://downloads.chef.io/chef-dk/.
```

You need to specify `--chefdk-version 1.5.0` to make things work.

See also the version number at [downloads.chef.io](https://downloads.chef.io/chefdk/stable/1.5.0).

Signed-off-by: tpetchel <tpetchel@gmail.com>


brewn:
Thanks @tpetchel!

PR #968, opened 2017-09-26
It seems we are missing a step

jjasghar:
Running the command i saw:

```
root@chefserver:~# chef-server-ctl set-secret data_collector token '93a49a4f2482c64126f7b6015e6b0f30284287ee4054ff8807fb63d9cbd1c506'
Please restart these services: nginx, opscode-erchef
```
This adds that command.


brewn:
Looks like this was added pretty recently, as I didn't see this message on my 12.15.8 server, but it shows up on 12.16.14. Thanks for catching this!

PR #971, opened 2017-09-27
reindex option note

sean-horn:
The -w option has restrictions on its use.


sean-horn:
I believe the failure in 2550837 above is remedied by adding the same change to the release notes file in the second commit. I didn't duplicate dtags entries like I should.

PR #973, opened 2017-09-27
Update to match Ruby 1.9+ Hash style

tomdoherty:
Signed-off-by: Tom Doherty <tom.doherty@fixnetix.com>


brewn:
Thanks @tomdoherty!

PR #979, opened 2017-09-28
Fix typo on the windows-package resource page

martinmosegaard:


brewn:
Good catch, thanks @martinmosegaard 
tenor-259905691

PR #986, opened 2017-10-02
Removed references to supports property from page.

jdgoins:
The `supports` property is deprecated and should no longer be included in this page.

### DIFF
--- a/chef_master/source/resource_service.rst
+++ b/chef_master/source/resource_service.rst
@@ -39,7 +39,6 @@ The full syntax for all of the properties that are available to the **service** 
      status_command             String
      stop_command               String
      subscribes                 # see description
-     supports                   Hash
      timeout                    Integer # Microsoft Windows only
      action                     Symbol # defaults to :nothing if not specified
    end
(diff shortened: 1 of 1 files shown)


jdgoins:
It seems as though this supports property is still useful for the service resource. I'm not aware of an equivalent property that fills that gap. I'm confused by the property being deprecated, but also still serving a purpose.

brewn:
Since this is still in use on Chef 12, I think for now it would be appropriate for us to add a warning note for the property - like what was done here: https://docs.chef.io/resource_common.html

jdgoins:
Yeah, I think that would probably make more sense. With that said, perhaps we should continue removing the property from most of the examples.

iennae:
Where are you getting the information that supports is deprecated for the service resource?

jdgoins:
I guess I'm confused by https://docs.chef.io/resource_common.html saying that supports is deprecated while the service resource uses the same property. Is it not deprecated for some resources?

iennae:
It was on all resources before but only implemented internally on the user and service resource. It was removed from the user and base class which is where the deprecation notices are. It's still implemented internally on the service resource.

jdgoins:
Hmmm okay. Thanks for clarifying this. I'd prefer it deprecated entirely to avoid confusion, but that's a different conversation. I'll close this PR.

PR #989, opened 2017-10-03
Updated to the new link for chefspec

jjasghar:
- Update link


brewn:
Good catch, thanks!

PR #993, opened 2017-10-04
Add list of all supported operating systems

jjasghar:
Taken from the chef-client cookbook, this is a list of all the supported oses.


brewn:
good stuff as always, thanks @jjasghar

PR #996, opened 2017-10-05
Fixed typos in elasticsearch config section

jmassardo:
A user in the slack channel was having issues and it was identified that the doc had a typo. Fixed two spots where elasticsearch was spelled wrong.


brewn:
Thanks @jmassardo! 
tenor-132768127

PR #997, opened 2017-10-06
Add note about data bag naming restrictions.

sandratiffin:
Please hold for 13.6. This note accompanies the fix for https://github.com/chef/chef/issues/3058.

Or, this update could be merged now; the restriction still applies as knife search will fail if any of the
listed names is used, it just isn't currently enforced ¯\_(ツ)_/¯.


brewn:
Not sure why this didn't close automatically when I merged #1071 ..but anyway, it's merged there. 

Thanks for adding this @sanditiffin

Issue #965, opened 2017-09-25
https://docs.chef.io/resource_registry_key.html outdated style


budgester:
Still uses old hash syntax.

e.g.

```
registry_key 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\TCPIP\Parameters' do
  values [{
    :name => 'MaxUserPort',
    :type => :dword,
    :data => 5000
  }]
  action :create
end
```

Gives the following in cookstyle.

```
Inspecting 11 files
........C..

Offenses:

recipes/windows_tools.rb:21:5: C: Use the new Ruby 1.9 hash syntax.
    :name => 'TcpTimedWaitDelay',
    ^^^^^^^^
recipes/windows_tools.rb:22:5: C: Use the new Ruby 1.9 hash syntax.
    :type => :dword,
    ^^^^^^^^
recipes/windows_tools.rb:23:5: C: Use the new Ruby 1.9 hash syntax.
    :data => 30
    ^^^^^^^^
recipes/windows_tools.rb:23:5: C: Put a comma after the last item of a multiline hash.
    :data => 30
    ^^^^^^^^^^^
recipes/windows_tools.rb:30:5: C: Use the new Ruby 1.9 hash syntax.
    :name => 'MaxUserPort',
    ^^^^^^^^
recipes/windows_tools.rb:31:5: C: Use the new Ruby 1.9 hash syntax.
    :type => :dword,
    ^^^^^^^^
recipes/windows_tools.rb:32:5: C: Use the new Ruby 1.9 hash syntax.
    :data => 5000
    ^^^^^^^^
recipes/windows_tools.rb:32:5: C: Put a comma after the last item of a multiline hash.
    :data => 5000
    ^^^^^^^^^^^^^

11 files inspected, 8 offenses detected
```

PR #1001, opened 2017-10-09
Updated registry resource to use ruby 1.9 hash syntax

jmassardo:
Updated the hash style to match new style as identified in issue #965.

Also moved `action :create` to bottom of resource block on lines 717 and 729 to maintain consistency among blocks.


jmassardo:
I found a number of tags that didn't match. I'm closing this request while I resolve all of them. I'll submit a new PR shortly.

PR #1012, opened 2017-10-13
Correct typo in databag docs

gliptak:


brewn:
Good catch, thanks @gliptak

PR #1015, opened 2017-10-13
added a block about tuning the chef server

modrake:
Added a block for increasing default request size allowed by the Chef server. Customers who use a larger compliance profiles, such as those that ship with the Automate server, will receive 413 errors on chef-client runs if they associate those profiles with the audit cookbook. These settings remediate that issue; the hope is that having this in the walkthrough and the error in a searchable place on our docs page, will make this issue easier to research and address for both our customers and our customer facing engineers.


brewn:
Thanks @modrake!

PR #1021, opened 2017-10-16
fixed broken link to compliance scan doc

modrake:
was missing ".html"


brewn:
Nice catch, thanks @modrake

Issue #61, opened 2016-05-16
Fully document installation of supermarket


vinyar:
Issue:
 1 - Installation should not exclusively rely on community cookbook(s), custom data_bag, etc..  and it should not be a 'black box'.
 2 - a user interested in understanding the Supermarket setup process is tasked with reverse engineering a community cookbook to understand what's happening.

Expectation:
Supermarket installation should be fully documented, and any user should be able to setup Supermarket by hand (in a similar fashion to chef-server, analytics, etc..). A link to community resource which aids in the task can be provided as a secondary route for installation.

Reference:
https://docs.chef.io/install_supermarket.html

vinyar:
sidebar:
"Sufficient disk space to meet project cookbook storage capacity..." should be expanded into a more meaningful number - maybe 10 gb per 1000 cookbooks at 10kb - 2mb each.

vinyar:
another sidebar - under s3 bucket it should be annotated that encrypted s3 buckets are presently not supported.

PR #1022, opened 2017-10-16
initial pass at non-cookbook supermarket install section

brewn:

### DIFF
--- a/chef_master/source/install_supermarket.rst
+++ b/chef_master/source/install_supermarket.rst
@@ -128,6 +128,7 @@ All of the keys under ``node['supermarket_omnibus']`` are written out as ``/etc/
    default['supermarket_omnibus']['chef_oauth2_app_id'] = '14dfcf186221781cff51eedd5ac1616'
    default['supermarket_omnibus']['chef_oauth2_secret'] = 'a49402219627cfa6318d58b13e90aca'
    default['supermarket_omnibus']['chef_oauth2_verify_ssl'] = false
+   default['supermarket_omnibus']['fqdn'] = 'supermarket.mycompany.com'
 
 On your workstation, generate a new cookbook using the ``chef`` command line interface:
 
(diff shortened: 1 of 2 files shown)


brewn:
#61

chef_master/source/install_supermarket.rst
```diff
@@ -124,7 +124,7 @@ All of the keys under ``node['supermarket_omnibus']`` are written out as ``/etc/
 
 .. code-block:: ruby
 
-   default['supermarket_omnibus']['chef_server_url'] = 'https://chefserver.mycompany.com'
+   default['supermarket_omnibus']['chef_server_url'] = 'https://chefserver.mycompany.com:443'
```
robbkidd, line 127: Not sure why this port would need to be specified as that's the default for `https://`. Is it to trigger in the reader "OH, we run SSL on a different port than that!" thoughts?

chef_master/source/install_supermarket.rst
```diff
@@ -291,6 +291,57 @@ When the bootstrap operation is finished, do the following:
...
+
+     .. code-block:: bash
+
+        rpm -Uvh /path/to/package/supermarket*.rpm
+
+#. Run the ``reconfigure`` command to complete the initial installation:
+
+   .. code-block:: none
+
+      sudo supermarket-ctl reconfigure
```
robbkidd, line 321: Why this initial reconfigure before creating `/etc/supermarket/supermarket.json`?

chef_master/source/install_supermarket.rst
```diff
@@ -291,6 +291,57 @@ When the bootstrap operation is finished, do the following:
...
+
+#. Create an ``/etc/supermarket/supermarket.json`` file and add the following information, substituting the values for each configuration option with the OAuth 2.0 client credentials that were created in the `previous section </install_supermarket.html#chef-identity>`__:
+
+   .. code-block:: ruby
+
+      {
+          "chef_server_url": "https://chefserver.mycompany.com:443",
+          "chef_oauth2_app_id": "0bad0f2eb04e935718e081fb71asdfec3681c81acb9968a8e1e32451d08b",
+          "chef_oauth2_secret": "17cf1141cc971a10ce307611beda7ffadstr4f1bc98d9f9ca76b9b127879",
+          "chef_oauth2_verify_ssl": false
```
robbkidd: This list leaves out `chef_identity_url` which is necessary to set in the latest stable version but that will quickly be fixed by what's coming in a release soon.

We should also add setting `fqdn`to these steps for when the private Supermarket is to serve itself up as a name that's different than the hostname of the server.

robbkidd (changes requested):

chef_master/source/install_supermarket.rst
brewn: This initial reconfigure creates the `/etc/supermarket` directory

robbkidd: OK.

PR #1027, opened 2017-10-17
Automate sudoers

sean-horn:
Automate should have not have tty specific entries, as this causes various parts of Automate to fail


chef_master/source/install_chef_automate.rst
```diff
@@ -145,6 +150,8 @@ If you don't have an existing Chef server installed and configured, the steps be
 Chef Server Installation
 ------------------------------------------------------
 
+.. note:: The standalone Chef server installation is not needed for most workflows. Only install the standalone Chef server if your workflow depends on it.
```
brewn: I'm confused by this note...the standalone process here is basically the same as the one in the Chef server guide:

https://docs.chef.io/install_server.html#standalone

I think standalone Chef servers are very common..

PR #1035, opened 2017-10-21
Document enforce_idempotency

gordonbondon:
`enforce_idempotency` was introduced here test-kitchen/test-kitchen#875 but never documented


brewn:
Thanks @gordonbondon!

Issue #561, opened 2017-04-05
Windows Scheduled Task install method lacks interval info


jblaine:
https://docs.chef.io/install_windows.html lacks information about how the scheduled task will be configured by default when the installing user chooses the Scheduled Task method

Issue #1034, opened 2017-10-21
Dangling ::note in config_rb_metadata


iennae:
Not entirely sure how to fix the problem here but wanted to note it:

https://github.com/chef/chef-web-docs/blob/master/chef_master/source/config_rb_metadata.rst

has a dangling note
![screen shot 2017-10-20 at 9 05 08 pm](https://user-images.githubusercontent.com/1592404/31847921-729cedd4-b5da-11e7-84f9-fa36777d8474.png)

PR #1036, opened 2017-10-22
fixed note element

jmassardo:
Fixed note element as identified in issue #1034. Added carriage returns before and after so element renders correctly.


jmassardo:
Added scheduled tasks option info from the windows page to the install_windows page to resolve issue #561

brewn:
Another great addition, thanks @jmassardo

PR #1050, opened 2017-10-24
update lazy example to be functional

spion06:
The current example given for how to use the lazy method for a variable into a template doesn't function unless you assume the template is aware it is being passed a proc object. This update makes it so it functions more as expected

### DIFF
--- a/chef_master/source/resource_common.rst
+++ b/chef_master/source/resource_common.rst
@@ -501,9 +501,9 @@ The following example shows how to use lazy evaluation with template variables:
 
    template '/tmp/canvey_island.txt' do
      source 'canvey_island.txt.erb'
-     variables(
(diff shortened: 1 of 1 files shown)


brewn:
@spion06 I ran this through cookstyle to update it to Ruby 1.9 hash syntax, which is what I've added here: #1063 

If this doesn't work for you, let me know, and I'll reach out to our engineers for clarification.

spion06:
unfortunately the problem isn't with the rocket style syntax. it's with the location of the lazy function. In the example the value of the key is a proc object. In order for delayed evaluation to function properly you have to do `@canvey_island.call` vs just doing `@canvey_island`. Perhaps this is intended and templates should be aware of this? I have never written a template like that and it seems awkward to.

PR #1057, opened 2017-10-26
Update the ohai-11 and ohai-12 deprecation timelines

tas50:
Also include a note about the foodcritic rules.


brewn:
Thanks, @tas50

PR #1059, opened 2017-10-28
Added -F to knife node show list, updated others

ctataryn:
Using knife `12.19.36` when I execute a `knife node show [--long | --medium] <node_name>` it does not display JSON. Instead it's formatted like this:

```
Environment: uat                                                       
FQDN:                                                                       
IP:                                                                         
Run List:    role[tomcat-server], role[tomcat-app], recipe[corp::filebeat]
Roles:                                                                      
Recipes:                                                                    
Platform:                                                                   
Tags:                                                                       
Attributes:                                                                 
app:                                                                        
  version: 1.0-SNAPSHOT                                                     
elk:                                                                        
  filebeat:                                                                 
    inputs:                                                                 
      name: app-log                                                        
      path: C:\tc-instances\app\logs\app.log                              
    outputs:                                                                
      host: bar.foo.com                                         
      port: 5044                                                            
      type: logstash                                                        
tags:                                                                       
```
It is only when I use the `-F json` flag do things display in JSON.

Also, I'm questioning the _New in Chef Client 12.0_ verbage, as I took at glance at v11.16 [1] where `--long` and `--medium` are listed there as well.

[1] - https://docs-archive.chef.io/release/11-16/knife_node.html

### DIFF
--- a/chef_master/source/knife_node.rst
+++ b/chef_master/source/knife_node.rst
@@ -654,13 +654,16 @@ This argument has the following options:
 ``-a ATTR``, ``--attribute ATTR``
    The attribute (or attributes) to show.
 
+``-F json``, ``--format=json``
+   Display output as JSON.
+   
 ``-l``, ``--long``
-   Display all attributes in the output and show the output as JSON.
+   Display all attributes in the output.
 
    New in Chef Client 12.0.
 
 ``-m``, ``--medium``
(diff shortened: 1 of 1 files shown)


brewn:
Good addition, thanks @ctataryn 

I'm going to look into the "New in.." note, but it appears you're right about that.

PR #1060, opened 2017-10-30
Added detail around CIS profiles

modrake:
Included a distinction between Windows Server and Desktop profiles, as this affects many users and prospects, and clarified that the CIS profiles include both levels 1&2.


kagarmoe:
tenor-245513420

PR #1079, opened 2017-11-01
Adding missing macOS 10.13 and Windows 2016

tas50:
We have buildings and testers for these now.


juliandunn (approved):
It's probably time to scrap 10.10 as well since Apple only actively supports 3 current versions

tas50:
I was trying to find a doc where apple spelled that out. Got a link?

tas50:
Apparently there is no doc, but it 10.10 updates have stopped. I'll cut another PR after this one is merged to fix that up too

PR #1104, opened 2017-11-13
Updating documentation to to include Ubuntu 16.04

Defilan:


brewn:
This change looks fine to me, but I think this table should be removed from the push jobs guide in favor of the platforms page.

juliandunn (approved):
Looks good to me, and I'm with you @brewn on DRYing it up, as the platforms page is authoritative for all support statements.

PR #1114, opened 2017-11-15
Kitchen Provisioner Setting Updates

wrightp:
Initial changes per [RFC091](https://github.com/chef/chef-rfc/blob/master/rfc091-deprecate-kitchen-settings.md)

Update existing settings with deprecation messages in descriptions.

Add `New Provision Settings` table.

Eventually when the new settings become the default settings the table will be unified and deprecated settings will be removed.

@chef/jex-team @coderanger @cheeseplus 

@coderanger We briefly discussed adding Kitchen deprecation warnings to the [Chef Deprecation Warnings](https://docs.chef.io/chef_deprecations_client.html) page. After seeing this change do you still think it will be necessary?

### DIFF
--- a/chef_master/source/config_yml_kitchen.rst
+++ b/chef_master/source/config_yml_kitchen.rst
@@ -115,6 +115,9 @@ This file uses Vagrant as the driver, which requires no additional configuration
 
 Provisioner Settings
 ==========================================================================
+Test Kitchen's provisioner settings will be changing in a future version. See `Chef RFC 091 <https://github.com/chef/chef-rfc/blob/master/rfc091-deprecate-kitchen-settings.md>`_
+for details. Settings that will be deprecated are listed in the descriptions below. The new recommended settings are listed in the `New Provisioner Settings` table.
+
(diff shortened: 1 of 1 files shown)


coderanger:
We should really remove this whole page in favor of something on kitchen.ci because having two is bad. If you want to point the links here from the doctor output, it's all the same to me as long as there is info somewhere :)

coderanger:
Also this needs way more examples of how to convert old -> new.

wrightp:
YES! I'm all for moving this page to kitchen.ci. I'll start working on that. Also 👍 on more examples.

brewn:
👍  on moving this to kitchen.ci

wrightp:
After spending time migrating the .kitchen.yml page content from chef-web-docs to kitchen-docs I think we need to consider migrating the content as a separate task.  It's more effort than I initially realized. Not just reformatting the markdown, but there are also design considerations.

I think we should merge this change to get the info out there. The community and docs teams can work together on what the best approach is on migration all the kitchen content so it's streamlined and consistent.

PR #1122, opened 2017-11-20
The guard's environment _is_ inherited from the parent resource.

### LINKED ISSUE #887: cwd of guard is inherited from its parent resource, even if it is an execute resource
In the note of the Guard section at https://docs.chef.io/resource_execute.html
we can read
`When using the not_if and only_if guards with the execute resource, the current working directory property (cwd) is not inherited from the resource.`

However
```shell
$ chef-shell
chef (12.21.4)> recipe_mode
chef: recipe (12.21.4) > execute 'cwd guard bug' do
chef:recipe > cwd 'not existing dir'
chef:recipe ?> command 'echo hello'
chef:recipe ?> not_if 'exit 0'
chef:recipe ?> end
...
chef:recipe (12.21.4)> run_chef
...
/home/j.alberdi/.gem/ruby/2.3.0/gems/mixlib-shellout-2.3.2/lib/mixlib/shellout/unix.rb:185:in

'chdir': No such file or directory @ dir_chdir - not existing dir (Errno::ENOENT)
...
```
This seems to have been implemented at https://github.com/chef/chef/pull/2223

If the issue is confirmed, updating the doc / reverting the feature would be welcome.

juliandunn:
fixes #887

### DIFF
--- a/chef_master/source/resource_execute.rst
+++ b/chef_master/source/resource_execute.rst
@@ -303,13 +303,13 @@ A guard property is useful for ensuring that a resource is idempotent by allowin
 
 .. note:: .. tag resources_common_guards_execute_resource
 
-          When using the ``not_if`` and ``only_if`` guards with the **execute** resource, the current working directory property (``cwd``) is **not** inherited from the resource. For example:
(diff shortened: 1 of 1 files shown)


brewn:
thanks @juliandunn

PR #1149, opened 2017-12-08
Round brackets for git resource environment attribute

chen01:
I found in testing that curly brackets in the environment attribute causes my chef runs to break. 
I tested using ruby versions 2.2.1 and 2.4.1

Note: There should be no space between `environment` and the open bracket. Having a space there also breaks too.


brewn:
Hi @chen01,

Thanks for bringing this to our attention - this definitely needed updating. After speaking to one of our engineers we determined that it shouldn't be necessary to use brackets here at all, as they're optional in ruby methods. I've amended this commit and merged the final product here: https://github.com/chef/chef-web-docs/pull/1152

This should work for you if you'd care to try it in a Chef run. Thanks again for catching this issue.

Issue #180, opened 2016-10-26
manage_symlink_source explanation


sasideveloper:
There is no documentation how chef client interprets when manage_symlink_source is set to false.

PR #1154, opened 2017-12-12
document "false" value for manage_symlink_source

brewn:

### DIFF
--- a/chef_master/source/resource_cookbook_file.rst
+++ b/chef_master/source/resource_cookbook_file.rst
@@ -42,7 +42,7 @@ The full syntax for all of the properties that are available to the **cookbook_f
      force_unlink               TrueClass, FalseClass
      group                      String, Integer
      inherits                   TrueClass, FalseClass
-     manage_symlink_source      TrueClass, FalseClass, NilClass
+     manage_symlink_source      TrueClass, FalseClass
      mode                       String, Integer
      notifies                   # see description
      owner                      String, Integer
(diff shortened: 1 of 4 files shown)


brewn:
I'm basing this on this comment: https://github.com/chef/chef/blob/db57131ad383076391b9df32d5e9989cfb312d58/spec/support/shared/functional/file_resource.rb#L409

brewn:
https://github.com/chef/chef-web-docs/issues/180

lamont-granquist:
The verbiage is still vaguely horrible.

Change the behavior of the file resource if it is pointed at a symlink. Possible values: ``true``, or ``false``. When this value is set to ``true``, the Chef client will manage the symlink's 
permissions or will replace the symlink with a normal file if the resource has content. When this value is set to ``false``, Chef will follow the symlink and will manage the permissions and content 
of the target file of the symlink. The default behavior is ``true`` but emits a warning that the default value will be changed to ``false`` in a future version, setting explicitly to ``true`` or 
``false`` suppresses this warning.

- the nil value is strictly an implementation detail of how the default is handled, users should not know or care about that, and it confuses the user and allows them to miss the fact that there's really only two fundamental behaviors here.
- phrasing like "manage a symlink's source file" is a confusing way to say "manage the symlink" (instead of the target file).   the name of the property is a bit of a confusing mess, and we need more clarity here rather than doubling-down on the confusing phrasing.
- the warning message / defaults changing needs to come last, and the behavior needs to be put in front of the users first.
- the behavior needs to have slightly better explanation -- if its `true` we either manage the symlinks perms or we clobber the symlink and replace it -- if its `false` we follow the symlink and manage the modes and content of the target file.  and there's two possible use cases for `manage_symlinik_source true`, one is that the user can manage the perms of the symlink, the other is that the symlink gets blown away with a normal file, and that needs to get spelled out (when a user comes here looking for "how do i manage the perms on a symlink?" they need to read that in a sentence clause -- if they come here looking for "how do i replace a symlink with a normal file?" they need to read that in a sentence clause -- the existing verbiage asks them to do verbal math in order to get their answer on words that are too abstract and confusing).

brewn:
@lamont-granquist thanks for clearing up what this does exactly, I agree that the existing verbiage is unclear..what you've written up is much better.

lamont-granquist:
👍

PR #1157, opened 2017-12-13
Add an example of using options with chocolatey

spuder:


brewn:
Another good addition, thanks @spuder

PR #1181, opened 2017-12-29
Remove reference to compat_resource

tas50:
We're trying to pretend we never wrote that


kagarmoe:
https://gph.is/XHIRfu

PR #1182, opened 2017-12-29
Add a missing new_resource and simplify examples

tas50:
Remove the default values for file/directory. This just adds noise and confuses


kagarmoe:
image

PR #1184, opened 2017-12-29
Fix load_current_value example to actually work

tas50:
You can't use new_resource here without some additional magic.


kagarmoe:
tenor-169026718

PR #1198, opened 2018-01-08
properly indent example knife plugin in auth.rst

adamvduke:
The example plugin on https://docs.chef.io/auth.html in the 'API Requests' section seems to be incorrectly indented.


kagarmoe:
tenor-158073501

PR #1194, opened 2018-01-05
[MSYS-744]Update doc for create & delete action

dheerajd-msys:
Updates the https://docs.chef.io/resource_windows_service.html  for action `create` & `delete`

PR #1199, opened 2018-01-10
Revert "[MSYS-744]Update doc for create & delete action"

brewn:
Reverts chef/chef-web-docs#1194


brewn:
This isn't in https://github.com/chef/chef yet so holding off on this till then.

PR #1250, opened 2018-02-16
cleanup example audit wrapper attribute formatting

mbainter:
The current example's quoted keys isn't valid syntax. Updated to correct this and adjusted some stylistic issues that will throw foodcritic/cookstyle errors (extra commas, indentation issues)


brewn:
Good catch, thanks @mbainter

PR #1252, opened 2018-02-20
Update powershell_package with source property

Happycoil:
https://github.com/chef/chef/pull/6843


brewn:
Thanks @Happycoil, we'll merge this when the next Client release goes out.

tas50:
This ships in 14.0 so we'll want to note that in the docs

Happycoil:
@tas50 Is there a standardized way of doing so? I was thinking a simple "Introduced in Chef Client 14" under the property description.

PR #1257, opened 2018-02-21
Document new family option in ifconfig resource

martinisoft:
Now that https://github.com/chef/chef/pull/6883 is merged I have created some documentation for the new option. I can add an example to the bottom of the examples section if needed, but I'd have to get some tips on marking it as a Debian specific thing.

I am also unsure which Chef release this will make it into so I have not marked which Chef Client version yet. I'm assuming Chef 13.7? If not, then 14.

### DIFF
--- a/chef_master/source/resource_ifconfig.rst
+++ b/chef_master/source/resource_ifconfig.rst
@@ -29,6 +29,7 @@ The full syntax for all of the properties that are available to the **ifconfig**
      bootproto                  String
      device                     String
      ethtool_opts               String
+     family                     String # defaults to 'inet' if not specified
      hwaddr                     String
      inet_addr                  String
      mask                       String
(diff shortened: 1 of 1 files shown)


brewn:
Thanks @martinisoft, I'll add the version specifics and merge once the next Client release goes out.

tas50:
This will go out in 14.0

chef_master/source/resource_ifconfig.rst
```diff
@@ -29,6 +29,7 @@ The full syntax for all of the properties that are available to the **ifconfig**
      bootproto                  String
      device                     String
      ethtool_opts               String
+     family                     String # defaults to 'inet' if not specified
```
tas50, line 32: We should leave the default for the description below

martinisoft: Oh, I did not catch this comment. It can be corrected after merge.

PR #1261, opened 2018-02-27
FIPS mode support

sean-horn:
FIPS mode is only supported by Chef Server and chef-client/chefDK
All other add-ons, like Reporting, Push Jobs and so on are not included.


sean-horn:
superceded by 1262

PR #1275, opened 2018-03-08
Use bento slugs in the kitchen.yml examples

tas50:
This entire page needs to be rewritten, but this is a step in the right direction


brewn (approved):
Feel free to open an issue for the other changes this page needs - I don't know that it's something I'll be able to get to in the near future, but we it'll be helpful to have a quick summary of what needs changing.

PR #1276, opened 2018-03-10
add debian and alphabetize EOL date table

brewn:


brewn:
After I did this I realized that the table was previous organized by EOL date - I found it a bit confusing to look through it that way, but I can undo the alphabetizing if that's what we want.

tas50:
Thanks

scotthain (approved):

PR #1278, opened 2018-03-13
Update property descriptions and remove provider section

tas50:
No one needs to know about providers other than developers
Some of the property descriptions were incorrect
Some of the defaults were incorrect
Lets just call the type True & False. It's technically not a type, but it makes a lot more sense to users


brewn:
Yeah, agreed re: trueclass / falseclass. Updating that here: https://github.com/chef/chef-web-docs/pull/1292

PR #1304, opened 2018-03-20
Simple prompt change

georgespatton:
Simple root prompt change.


brewn:
Thanks @georgespatton

PR #1309, opened 2018-03-23
Update integrate_delivery_ldap.rst

acranox:
The default for ldap_bind_dn appears to be `ldapbind`, and leaving this and the password values unset will cause automate to use the default values.  Setting them to an empty string creates the desired behavior for an unauthenticated bind.


kagarmoe:
Thanks! We now discuss this problem in line 41.

PR #1323, opened 2018-04-02
Update install_omnibus.rst

username-is-already-taken2:
Corrected the URL Link as using https returned a forbidden message


brewn:
Not really sure why that would happen - I can confirm the https link works for me. How are you pulling the info, from `curl`, `wget`, or your browser, etc?

brewn:
Probably related to this issue:

https://github.com/chef/omnitruck/issues/229

I'm going to follow up on that - we want to encourage use of https in the docs, but this needs to be sorted out elsewhere.

Thanks for the heads-up!

PR #1355, opened 2018-04-11
Updated documents for MSYS-752 changes.

Vasu1105:
Updated docs based on changes made in [#6815](https://github.com/chef/chef/pull/6815)

Signed-off-by: Vasu1105 <vasundhara.jagdale@msystechnologies.com>

### DIFF
--- a/chef_master/source/resource_windows_task.rst
+++ b/chef_master/source/resource_windows_task.rst
@@ -5,7 +5,7 @@ windows_task
 
 Use the **windows_task** resource to create, delete or run a Windows scheduled task. Requires Windows Server 2008 or later due to API usage.
 
-**New in Chef Client 13.**
+**New in Chef Client 14.1**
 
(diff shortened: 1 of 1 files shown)


Vasu1105:
Merge after [#6815](https://github.com/chef/chef/pull/6815) gets released.

tas50:
We need to update this to reflect that the individual properties are now in 14.1, not the whole resource. We still support Chef 13 so users need to understand what works in 13.X and what works in 14.1.

brewn:
Yeah, I'll fix the versioning stuff before this goes out.

Vasu1105:
@tas50 All of the properties are same. I am checking where I specifically mention about version 14.1. I have added more elaborate description about the properties so that user can use it wisely.

btm:
Chef 14.1 was released today. https://discourse.chef.io/t/chef-client-14-1-released/13032

tas50:
@brewn We should be good to merge this now

PR #1363, opened 2018-04-12
Enable expeditor and pipeline

scotthain:

This is the first step in upgrading the chef-web-docs internal deploy process!

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -1,0 +1,9 @@
+slack:
+  notify_channel: releng-notify
+
+github:
+  delete_branch_on_merge: true
+
+pipelines:
+  - verify:
+      description: Pull Request validation tests
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -1,0 +1,7 @@
+- label: "build"
+  command:
+    - pip install -r requirements
+    - make docs
+  plugins:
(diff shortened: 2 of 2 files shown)


.expeditor/verify.pipeline.yml
```diff
@@ -0,0 +1,7 @@
+- label: "[chef/chef-web-docs] build"
```
tduffield: the `[]` bit is unnecessary

.expeditor/config.yml
```diff
@@ -0,0 +1,6 @@
+slack:
+  notify_channel: releng-notify
+
```
schisamo, line 3: You'll want to add:

```yaml
github:
  delete_branch_on_merge: true
```

schisamo (changes requested):

.expeditor/verify.pipeline.yml
```diff
@@ -0,0 +1,7 @@
...
+    command:
```
tduffield: this should all be at the same indent level as label

tduffield (changes requested):

.expeditor/config.yml
scotthain: Done

schisamo (approved):

tduffield (approved):

PR #1365, opened 2018-04-13
clarify the updated deprecation

lamont-granquist:
you might also want to include a pointer to the discussion on `updated_by_last_action?` in https://github.com/chef/chef-web-docs/blob/master/chef_master/source/custom_resources_notes.rst


lamont-granquist:
👍

PR #1367, opened 2018-04-16
Typo Fix in doc `Workstation.rst` <an be downloaded> -> <can be downloaded> 

xysr89:
`an be downloaded` -> `can be downloaded`

tried to follow https://github.com/chef/chef-web-docs#the-fastest-way-to-contribute but not sure where the button **Propose file change** is. since its a small change i assume it may not be needed


brewn:
Thanks @xysr89

PR #1364, opened 2018-04-12
Create basic verification pipeline

scotthain:
This sets up the build step for testing

PR #1369, opened 2018-04-16
Setup build pipeline with terraform

scotthain:
@chef/jex-team 

supersedes https://github.com/chef/chef-web-docs/pull/1364

### DIFF
--- a/.expeditor/buildkite/deploy_docs.sh
+++ b/.expeditor/buildkite/deploy_docs.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -eou pipefail
+
+aws-configure chef-cd
+
+ceval "aws s3 sync build s3://chef-web-docs-${ENVIRONMENT}.cd.chef.co --profile chef-cd"
--- a/.expeditor/deploy.pipeline.yml
+++ b/.expeditor/deploy.pipeline.yml
@@ -1,0 +1,24 @@
+steps:
+  - label: "Terraform"
+    command: .expeditor/buildkite/terraform.sh
+    concurrency: 1
+    plugins:
+      docker#v1.1.1:
+        image: "chefes/buildkite"
+        environment:
+          - CHEF_CD_AWS_ACCESS_KEY_ID
+          - CHEF_CD_AWS_SECRET_ACCESS_KEY
+
+  - wait
+
+  - label: "Deploy"
+    command:
+      - .expeditor/buildkite/build_docs.sh
+      - .expeditor/buildkite/deploy_docs.sh
+    concurrency: 1
+    plugins:
+      docker#v1.1.1:
+        image: "chefes/buildkite"
+        environment:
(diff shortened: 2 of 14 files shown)


.expeditor/buildkite/deploy_docs.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -eou pipefail
+
+aws-configure chef-cd
+
+ceval "aws s3 sync build s3://chef-web-docs-${acceptance}.cd.chef.co --profile chef-cd"
```
tduffield: you aren't setting the variable `acceptance` - did you mean for this to just be a full string?

.expeditor/deploy.pipeline.yml
```diff
@@ -0,0 +1,24 @@
+steps:
+  - label: "Terraform"
+    command: .expeditor/buildkite/terraform.sh
+    concurrency: 1
```
tduffield, line 4: You need to specify a `concurrency_group` in both steps

Makefile
```diff
@@ -2,12 +2,18 @@ BUILDDIR = build
 BUILD_COMMAND = sphinx-build -a -W
 BUILD_COMMAND_AND_ARGS = $(BUILD_COMMAND)
 
-docs:
+# if we are in CI we need to add python's binpath:
+export PATH := $(shell hab pkg path core/python2)/bin:$(PATH)
```
tduffield: Can you clarify how/why this is only evaluated in CI and not locally?

terraform/s3.tf
```diff
@@ -0,0 +1,9 @@
+#########################################################################
+# Non production chef-web-docs s3 static site bucket
+#########################################################################
+
+module "chef-web-docs" {
+  source      = "git@github.com:chef/es-terraform.git//modules/cd_s3_website"
+  subdomain   = "chef-web-docs${var.environment == "delivered" ? "" : "-${var.environment}"}"
+  tag_contact = "shain"
```
tduffield: we probably want to make this `releng`

tduffield (changes requested):
Just a few minor things, otherwise looks good!

.expeditor/deploy.pipeline.yml
scotthain: is there a default or do I need to create one somewhere?

Makefile
scotthain: This is a misleading comment - we just need to add python to the path.

.expeditor/buildkite/deploy_docs.sh
scotthain: Copy paste error - should have been ENVIRONMENT

.expeditor/deploy.pipeline.yml
tduffield: you just have to specify it like so: https://github.com/chef/a2/blob/master/.expeditor/deploy.pipeline.yml#L5

PR #1384, opened 2018-04-18
Added recommendation for resource name

easydevmixin:
In Ubuntu if name has spaces it won't be taken into account when doing apth-get update.

E.g.:

``apt_repository "A Repo"...``

Will become:

``/etc/apt/sources.list.d/A Repo.list``

Which won't load when ``apt-get update`` is ran.


brewn:
Thanks @easydevmixin - I want to to a bit of testing on this before I copy edit / merge. Should be able to get this added soon.

PR #1403, opened 2018-05-02
list chef 13 as deprecated

brewn:

In keeping with: https://www.chef.io/eol-chef12-and-chefdk1/


tas50:
Is deprecated the work we're using for the N-1 release?

brewn:
@tas50 I would say yes, at least based off of that page. I think it's a bit confusing. But this change is just intended to keep the messaging consistent.

PR #1409, opened 2018-05-08
Add back Windows 2008R2 support, Add Ubuntu 18.04 and clarify SLES 12

tas50:
This should match reality now


stuartpreston (commented):
This is a nit but can we update this to have consistent style for Server and Version i.e. '2008 R2', '2012 R2'?  I also have thoughts on supported architectures but this can follow in a separate PR.

PR #1410, opened 2018-05-08
Consistent naming of Windows server releases

tas50:


brewn:
In general I see these versions used with a space between them, ie. `2012 R2`. Is there a reason to omit the space?

stuartpreston (approved):

stuartpreston (changes requested):
Can we just keep the uppercase version with the spaces (i.e. '2008 R2', '2012 R2')

stuartpreston (approved):

PR #1413, opened 2018-05-09
Updating the supported versions to deprecate Reporting

krasnow:
Marking Reporting as Deprecated. At the end of May I'll change the status of Reporting, Analytics, and Enterprise Chef to EOL with a date of December 31, 2018.


brewn:
If we know what the EOL date is going to be, why not include it in that column?

PR #1426, opened 2018-05-14
Automate disk space protection

sean-horn:
Lately, the Ohai Sessions plugin has been causing no end of trouble in Automate installations.
Let's just go ahead and make sure people know it's going to be a potential problem so that they can address it as they get started with sending node data to their new Automate system.


kagarmoe:
tenor-68646100

PR #1428, opened 2018-05-16
Remove outdated metadata fields

thommay:
RFC 85 removes the `recommends`, `suggests`, `conflicts` and `replaces` fields as they're not implemented and are confusing. We shouldn't document them.


tas50:
This is a great addition. People continue to use these even though we never implemented them. They're one of those things that just confuses new users and makes Chef harder to understand.

PR #1429, opened 2018-05-17
Attempt to stop overloading the word "name"

tas50:
It's causing confusion


tas50:
@brewn This came up in community chat today. This is step 1 in helping out beginners in these docs

PR #1431, opened 2018-05-22
Changing ubuntu versions at 'supports' as 12.04 is EOL

pschaumburg:


tas50 (approved):
We shouldn't use EOL distros in an example

kagarmoe:
tenor-4497689

PR #1436, opened 2018-05-29
Property document the sensitive: true functionality

tas50:
The brief mention at the top of the page isn't super helpful.


kagarmoe:
Thanks!

PR #1438, opened 2018-05-30
Remove mentions of ChefDK 1.2

tas50:
We support ChefDK 2 and 3 at this point. We can assume everyone is on 1.2 or later now


kagarmoe:
tenor-133097181

PR #1439, opened 2018-05-30
Remove metadata.rb documentation regarding chef 12 and attributes

tas50:
If we introduced it in early Chef 12 releases we really don't need to mention that anymore. It's fair game for everyone. Also remove the attributes stuff which no one should ever use.


kagarmoe:
tenor-238435615

PR #1441, opened 2018-05-30
Move the audit DSL to its own doc

tas50:
control and control_group are from the old audit-mode, which we shouldn't be introducing to users who are trying to learn the Chef DSL. It just complicates things and makes it that much harder to wrap your head around Chef. Toss them in their own doc that we can eventually delete when audit-mode goes away.


kagarmoe:
tenor-238918

PR #1443, opened 2018-05-30
Remove an old first edition book and remove analytics tags

tas50:
We had the same book in this list twice. We also had analytics tag for packt. We shouldn't be tracking our users.


kagarmoe:
tenor-74493702

PR #1446, opened 2018-05-31
Add oracle to the RHEL list for platform_family

tas50:
There's far more people using oracle than scientific


kagarmoe:
tenor-9652992

PR #1450, opened 2018-06-01
Call it the Chef install script

tas50:
If you're not collecting a paycheck from Chef you probably don't care that we call it the Omnitruck script. Just call it the Chef install script.

### DIFF
--- a/chef_master/source/install_chef_air_gap.rst
+++ b/chef_master/source/install_chef_air_gap.rst
@@ -170,7 +170,7 @@ Install Chef DK
 
 Create a bootstrap template
 -----------------------------------------------------
-By default, ``knife bootstrap`` uses the ``chef-full`` template to bootstrap a node. This template contains a number of useful features, but it also attempts to pull an installation script from ``omnitruck.chef.io``. In this section, you'll copy the contents of the ``chef-full`` template to a custom template, and then modify the package and Ruby gem sources.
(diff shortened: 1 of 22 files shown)


schisamo:
We'll need to update references at https://docs.chef.io/packages.html#omnitruck-install-scripts also.

kagarmoe (commented):
I like this change, but to be consistent we will need to make in in 23 different documents. Should we consider changing the page & url to `install_chef` ?

chef_master/source/install_chef_air_gap.rst
```diff
@@ -66,7 +66,7 @@ In this section you'll install the Chef server, and create your organization and
 
 .. note:: If you intend to use Chef Automate, create the ``delivery`` user and add it to your organization during this step.
 
-#. Download the package from https://downloads.chef.io/chef-server/.
+#. Download the package from https://omnitruck.chef.io/chef-server/.
```
tas50: This should be https://downloads.chef.io/chef-server/

chef_master/source/install_server_ha.rst
```diff
@@ -454,10 +454,10 @@ The ``chef-backend-ctl gen-server-config`` command, which can be run as root fro
 Software Versions
 ----------------------------------------------------------------
 
-The backend HA cluster uses the omnibus installer (https://github.com/chef/omnibus) to package all of the software
+The backend HA cluster uses the Chef installer (https://github.com/chef/omnibus) to package all of the software
```
tas50: We can remove that Github link

PR #1451, opened 2018-06-01
Update metadata documentation

tas50:
Remove some old notes mentioning things were new in super old chef 12 releases
Remove the attributes documentation since no one should ever use this
Clarify that metadata is used by the client and server


kagarmoe:
tenor-78622605

PR #1454, opened 2018-06-01
There is no 'rhel' platform

tas50:
That's a platform family. These docs are all kinds of screwed up. This is the first part in fixing it.


kagarmoe (commented):
Seems good to me. We have 81 uses of RHEL in 36 files. Mostly it looks like `platform` or `platform_family`.

tas50:
platform_family is correct.

kagarmoe:
tenor-85505017

PR #1435, opened 2018-05-29
Documented default_source :artifactory

StuporHero:
I have an artifactory instance that I need to use in my build pipeline. I am using Policyfiles, and in my research, I discovered [this Pull Request](https://github.com/chef/chef-dk/pull/1252) on the chef-dk repository. I was able to get it to work and decided to help the next guy who needed this by adding it to the documentation. I've already done the `dtags replicate` to propagate my change to everywhere it is needed.

The changes in the dtags markdown files was just some formatting that was brought to my attention by markdownlint.

Thanks!

PR #1455, opened 2018-06-02
Corrected policyfile typo #1435

StuporHero:
In #1435, I forgot to close a pair of backticks, but since I made sure I did everything else correctly, it slipped past us all until I noticed it in the live docs (which I had to refer to since I already forgot how to use the feature I documented). This one should be equally smooth with the added bonus of being correct!


kagarmoe:
tenor-100849767

PR #1456, opened 2018-06-03
Another omnitruck -> chef

tas50:
Less confusion for those outside the company


tas50:
I think this is a dupe now

PR #1457, opened 2018-06-05
Update env example to pass linting

DWSR:
Obvious fix

### DIFF
--- a/chef_master/source/resource_execute.rst
+++ b/chef_master/source/resource_execute.rst
@@ -399,7 +399,7 @@ The following examples demonstrate various approaches for using resources in rec
      command 'slapadd < /tmp/something.ldif'
      creates '/var/lib/slapd/uid.bdb'
      action :run
-     environment ({'HOME' => '/home/myhome'})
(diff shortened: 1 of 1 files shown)


chef_master/source/resource_execute.rst
```diff
@@ -399,7 +399,7 @@ The following examples demonstrate various approaches for using resources in rec
      command 'slapadd < /tmp/something.ldif'
      creates '/var/lib/slapd/uid.bdb'
      action :run
-     environment ({'HOME' => '/home/myhome'})
+     environment('HOME' => '/home/myhome')
```
coderanger, line 402: Should probably skip the parens entirely since we leave them off all other properties.

DWSR: I find this to be a bit more readable and explicit, especially if you break out into multiline, like this:
```ruby
environment(
  HOME: '/home/coderanger',
  PATH: '/usr/bin:/usr/bin/local'
)
```

kagarmoe:
Sorry, I can't fix the dtags problem from this repo.

PR #1465, opened 2018-06-13
Simplify log level property

tas50:


kagarmoe:
tenor-148719160

tas50 (commented):
There’s a double period now

PR #1477, opened 2018-06-21
The `new_resource.property` syntax isn't used in part of the docs

bmiller08:
Its used later, just not in the beginning examples.


kagarmoe:
tenor-134516320

PR #1482, opened 2018-06-26
update "lazy" example

sjmiller609:
The existing "lazy" example for template variables does not work - it passes the lazy object to the template rather than evaluating lazily and passing the result to the template.


tas50 (approved):

kagarmoe:
tenor-203225916

PR #1487, opened 2018-07-04
[MSYS-814] - Doc update for cron resource

Nimesh-Msys:
Delete action can only be performed on cron entries created by chef.
As a part of workaround, we may use other resources like `execute`.

Signed-off-by: [Nimesh](<nimesh.patni@msystechnologies.com>)


kagarmoe:
tenor-214713470

PR #1489, opened 2018-07-06
Add Example of Proxy Settings in .kitchen.yml

zicklag:
Added an example of how to load proxy settings from the local
environment in the `.kitchen.yml` file.


kagarmoe:
tenor-46727042

oops. rebase fixed.

PR #1494, opened 2018-07-11
Update example versions and distros

tas50:
Update this to the current state of things.


kagarmoe (approved):
We could start using variables for versions & distros.

PR #1498, opened 2018-07-12
Update resource_powershell_script.rst

stuartpreston:
As per https://github.com/chef/chef/blob/master/lib/chef/provider/powershell_script.rb#L128-L134

Signed-off-by: Stuart Preston <stuart@chef.io>


kagarmoe:
tenor-54024091

PR #1503, opened 2018-07-13
Update resource_windows_service.rst

stuartpreston:
Windows does not support :reload as an action to windows_service resource hence updating the docs to say so!


kagarmoe:
tenor-35541473

PR #1514, opened 2018-07-19
Kick the books down the list on the community docs

tas50:
Chef books are pretty out of date and we shouldn't emphasize them as the best way to get started.


kagarmoe:
tenor-61137931

PR #1515, opened 2018-07-19
Update our platform support for Amazon Linux 2.0

tas50:
"current version" doesn't mean much now that they have 2 versions. In reality we support all version of 1.x aka 2013-2018 and 2.0+


kagarmoe:
tenor-50716876

PR #1516, opened 2018-07-19
Update FreeBSD 10 EOL date, update SLES support, remove .x versions

tas50:
FreeBSD 10 EOL has been pushed out to the end of Oct. This updates our date. It also clarified we support SLES 12 SP1 or later. We had that in a few places, but not all. Finally it simplifies how we refer to major release versions.

### DIFF
--- a/chef_master/source/install_server_pre.rst
+++ b/chef_master/source/install_server_pre.rst
@@ -20,16 +20,16 @@ The following table lists the commercially-supported platforms and versions for 
      - Version
    * - CentOS
      - ``x86_64``
-     - ``6.x``, ``7.x``
+     - ``6``, ``7``
    * - Oracle Enterprise Linux
      - ``x86_64``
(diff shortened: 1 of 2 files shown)


juliandunn (commented):
LGTM but do we support SLES15 at all?

tas50:
SP5?

I also noticed on the FreeBSD website that their EOL is now in Oct not April. Not sure if we want to kick 10.4 out to Oct then

tas50:
Oh actual 15.0 We've never tested on that so who knows.

kagarmoe:
So, what have we decided on FreeBSD 10.4 and SLES 15.0?

tas50:
I think we need to actually update FreeBSD to support 10 until the end of Oct to match the vendor. SLES can stay as is until we test things

kagarmoe:
tenor-43036822

PR #1526, opened 2018-07-26
Property `freeze` changed to `freeze_on_upload` in `chef_mirror` resource

aabognah:
### Description

Since Cheffish v5 this deprecation warning (and later fail) appears when trying to use freeze:
"Property `freeze` on the `chef_mirror` resource has changed to `freeze_on_upload`." \
"Please use `freeze_on_upload` instead. This will raise an exception in a future version of the cheffish gem."
[Please describe what this change achieves]

### Definition of Done
replace `freeze` with `freeze_on_upload`

### Issues Resolved
the documentation is incorrect

### DIFF
--- a/chef_master/source/_templates/nav-docs.html
+++ b/chef_master/source/_templates/nav-docs.html
@@ -765,6 +765,11 @@
                 "url": "/resource_chef_handler.html"
               },
               {
+                "title": "chocolatey_config",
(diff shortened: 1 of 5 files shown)


tas50 (approved):

kagarmoe:
tenor-248023384

PR #1528, opened 2018-07-26
Execute deployments with Buildkite

schisamo:

### DIFF
--- a/terraform/vars/delivered.tfvars
+++ b/terraform/vars/delivered.tfvars
@@ -1,0 +1,5 @@
+vpc = "delivered"
+
+www_dns = "chef-web-docs-production"
+
+# fastly_fqdn = "docs.chef.io"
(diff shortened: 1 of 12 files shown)


scotthain (approved):

terraform/vars/delivered.tfvars
```diff
@@ -0,0 +1,5 @@
+vpc = "delivered"
+
+www_dns = "chef-web-docs-production"
+
+fastly_fqdn = "docs.chef.io"
```
tduffield: We have to be careful not to promote to production while this is in place. Maybe we comment it out?

tduffield (changes requested):

terraform/vars/delivered.tfvars
schisamo: Good call, will do that now.

tduffield (approved):

PR #1537, opened 2018-07-30
de-crufts config/redirects + code

kagarmoe:

### Description
Reduces /config/redirects.json from 4,903 to 779 redirects.

### Check List

- [ ] Spell Check
- [X] Local build
- [ ] Examine the local build
- [ ] All tests pass


doctools/clean_redirects.rb
```diff
@@ -0,0 +1,28 @@
+## written by Jerry Aldrich 
+
+require 'json'
+
+REDIRECTS_FILE = ARGV[0]
+DOCS_DIRECTORY = '/home/jerry/Projects/github.com/chef/chef-web-docs'
```
schisamo: We shouldn't hardcode any user paths here. I would recommend just skipping this constant and making line 10:

```ruby
rst_files = File.expand_path('../../chef_master/source/*.rst', __FILE__)
```

More details on the magic of `__FILE__` and `File.expand_path` at:
https://www.rubydoc.info/stdlib/core/File.expand_path
https://stackoverflow.com/questions/224379/what-does-file-mean-in-ruby

schisamo (changes requested):

schisamo (approved):

PR #1577, opened 2018-08-31
Properly name the yum_package page so we can autogenerate all resources

tas50:
All the filenames align with the resource names except yum. This fixes that and uses a redirect for anyone that bookmarked the old URL.


mjingle (approved):
A well-needed fix! Thanks!

PR #1579, opened 2018-08-31
Minor cleanup for the docs-builder cookbook

tas50:
I'm not sure if this is used anymore, but this removes a Chef 11-ism from the code and makes sure were testing on the latest CentOS bento box.


tas50:
We just nuked this instead. Much better idea.

PR #1596, opened 2018-09-05
Replace knife.rb with config.rb in docs

tas50:
With Chef 12 we renamed knife.rb to config.rb...we just didn't tell anyone. This PR updates our docs to refer exclusively to the config file as config.rb. It's mentioned all over the place. It's a pretty huge update.


lamont-granquist:
:+1:

kagarmoe (approved):

PR #1597, opened 2018-09-05
Add missing redirect for the old knife.rb page

tas50:
I missed this and the current search links point to it.


kagarmoe (approved):

kagarmoe:
Good catch!

PR #1612, opened 2018-09-09
Remove blurb of random changelog stuff

tas50:
We mention the introduced versions for each of these below.


tas50:
I ended up fixing more.

PR #1613, opened 2018-09-09
Cleanup the client config docs

tas50:
Remove the wall of text that includes just some of the new configs. We mention each one as necessary below

Remove the really old "Changed in" messaging that doesn't actually say what the value was before. Also Chef 12.0 was 4 years ago. It's been long enough ago that we can remove this.

Remove the odd message about audit mode going stable. That doesn't belong here at all.

Remove the backwards compatibility message about compile_time that is for users < 12.1. No need for this anymore.


mjingle (approved):
tenor-176747766

PR #1614, opened 2018-09-09
Add notes about encrypted databags v3 to release notes

tas50:
Add a note for 12.0 when we introduced the format
Add a note for 13.0 when we made it the default

This probably needs some copy magic.


mjingle (approved):
tenor-83082230

PR #1619, opened 2018-09-10
Remove Ruby 1.9 mention and Chef as a windows service

tas50:
We require Ruby 2.4+ now and it bumps a version every year so lets just remove this. Also we've removed running Chef as a service on Windows so remove the mention of this.


mjingle (approved):
tenor-50533631

PR #1621, opened 2018-09-11
Remove note about a now deprecated ChefDK release

tas50:
We've actually changed this all a few times.


mjingle (approved):
tenor-113521617

PR #1623, opened 2018-09-11
Remove broken link to chef verify

tas50:
Chef verify is for devs only and we removed it, but didn't get rid of this nav link.


mjingle (approved):
tenor-85458292

PR #1627, opened 2018-09-11
Don't document the config for a removed knife command

tas50:
We yanked this command entirely. No need to document how to config it.


mjingle (approved):
tenor-111213759

PR #1638, opened 2018-09-18
Rework how we generate chef repos a bit

tas50:
I think this is going to need some further refinement, but the technical aspect of it is correct now. The command is correct and the "manual" bit is all gone since this is not manual.

### DIFF
--- a/chef_master/source/install_chef_air_gap.rst
+++ b/chef_master/source/install_chef_air_gap.rst
@@ -148,13 +148,13 @@ Install Chef DK
 
    .. code-block:: bash
 
-      dpkg -i chefdk_2.0.28-1_amd64.deb
+      dpkg -i chefdk_3.2.30-1_amd64.deb
 
-#. Use the ``chef generate app`` command to generate your Chef repo:
+#. Use the ``chef generate repo`` command to generate your Chef repo:
 
    .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


chef_master/source/install_chef_air_gap.rst
```diff
@@ -148,13 +148,13 @@ Install Chef DK
 
    .. code-block:: bash
 
-      dpkg -i chefdk_2.0.28-1_amd64.deb
+      dpkg -i chefdk_3.2.30-1_amd64.deb
```
kagarmoe, line 151: Will this DK version work with A1?

tas50: Shouldn't be an issue. Also if we tell them to go to downloads they're going to get a version similar to this. Mostly I just try to update example chef or DK versions when I see them. Docs that mention apps versions from years ago tells me the docs are potentially outdated as well.

tas50:
LGTM. Thanks for the fixes @kagarmoe

mjingle:
FYI @tas50 - Headers can't jump from H1 to H3 in these Sphinx documents. You can have all the H1 headers you want, but there must be a H1 -> H2 -> H3 progression with no jumps.

mjingle (approved):
tenor-76947173

PR #1639, opened 2018-09-18
More resource updates from automation

tas50:
Bring in more of the automation corrections. Eventually we'll be 100%, but has to be done 1 reviewed bit at a time.


mjingle (approved):
tenor-142544541

PR #1642, opened 2018-09-20
Update more resource descriptions based on automation

tas50:

### DIFF
--- a/chef_master/source/resource_windows_service.rst
+++ b/chef_master/source/resource_windows_service.rst
@@ -215,9 +215,9 @@ The windows_service resource has the following properties:
    The number of times to catch exceptions and retry the resource.
 
 ``retry_delay``
-   **Ruby Type:** Integer
+   **Ruby Type:** Integer | **Default Value:** ``2``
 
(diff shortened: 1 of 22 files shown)


chef_master/source/resource_windows_service.rst
```diff
@@ -144,14 +144,14 @@ The windows_service resource has the following properties:
...
 ``ignore_failure``
-   **Ruby Types:** True, False
+   **Ruby Types:** true, false | **Default Value:** ``false``
 
-   Continue running a recipe if a resource fails for any reason. Default value: ``false``.
+   Continue running a recipe if a resource fails for any reason.
 
 ``init_command``
-   **Ruby Type:** String
+   **Ruby Type:** String | **Default Value:** ``nil``
```
tas50: If it's nil we should actually skip the default altogether. Nil is basically no default

mjingle: There are a few of those ``nil``s around. Will make a note of it and fix it.

mjingle (approved):
tenor-36303934

PR #1650, opened 2018-09-20
Don't call knife windows a plugin

tas50:
While technically true it ships in DK so no one outside Chef needs to know it's a plugin.


chef_master/source/install_windows.rst
```diff
@@ -7,7 +7,7 @@ Install the chef-client on Microsoft Windows
 
 The chef-client can be installed on machines running Microsoft Windows in the following ways:
 
-* By using the `knife windows </plugin_knife_windows.html>`__ plugin to bootstrap the chef-client; this process requires the target node be available via SSH (port 22) or by using the HTTP or HTTPS ports that are required by WinRM
+* By using the `knife windows </plugin_knife_windows.html>`__ to bootstrap the chef-client; this process requires the target node be available via SSH (port 22) or by using the HTTP or HTTPS ports that are required by WinRM
```
stuartpreston: Grammar: We should also remove the "the" before the link too.

stuartpreston: Replace: "via SSH (port 22) or by using the HTTP or HTTPS ports that are required by WinRM" with "via the WinRM port (typically port 5985)" or similar.

stuartpreston (changes requested):
Also, we provide a Chocolatey package that should be documented: https://chocolatey.org/packages/chef-client

tas50:
I totally missed that SSH reference. Thanks @stuartpreston

stuartpreston (approved):

tas50:
@mjingle Can we merge this one?

mjingle:
@tas50 We will. I didn't know if Stuart's comment about documenting a Chocolatey package was included on this PR or not. Making a card to follow-up on that later.

PR #1651, opened 2018-09-20
Remove the option to install chef-client as a windows service

tas50:
This isn't a thing anymore and isn't a thing in any supported release of Chef.


mjingle (approved):
tenor-72339795

PR #1654, opened 2018-09-20
Kill off chef provisioning vagrant and remove an odd warning

tas50:
Chef provisioning vagrant is entirely deprecated and we don't want people to find it. Remove the mentions throughout. Also remove other mentions of drivers that we don't support anymore. Also remove the odd warning here that wasn't really a warning.


mjingle (approved):
tenor-84115020

PR #1655, opened 2018-09-20
Simplify the libraries page

tas50:
Remove more advanced concepts. If you know Ruby you already know all the horrible things you can do. Let's explain to new users why they might want a library.


tas50:
This probably need some copy love, but the key change here is not starting off with the most extreme thing you can do with a library. Start people off with helpers. Everyone loves helpers.

mjingle (approved):
tenor-185492196

LGTM!

PR #1659, opened 2018-09-21
Remove redundant Ruby documenation

tas50:
Move the log stuff out of the Ruby section and into the DSL section since it's Chef specific. Remove duplicate Ruby docs from the DSL section. Keep everything describing Ruby in a single Ruby doc.


mjingle (approved):
tenor-47970385

PR #1663, opened 2018-09-21
Remove Chef Provisioning from the nav bar and other docs

tas50:
Remove the nav bar / index entries for Chef provisioning. Also remove a few bits where we mentioned it in other docs. Most of this information was dated and wrong anyways. The world is a better place with this stuff hidden.


mjingle (approved):
tenor-14665507

PR #1664, opened 2018-09-21
Remove references to the old old web UI

tas50:
Manage is the old UI. The old web UI is the old old one.


mjingle (approved):
tenor-181079272

PR #1666, opened 2018-09-21
Move around some Chef items in the nav bar

tas50:
This is similar to the concepts cleanup stuff, but a bit less intense and also not really related to concepts. This is more about cleanup up the getting started bits.

- Move Release notes from Getting Started to the top level Chef section. I struggle to find these every time I hit the site. They're buried.
- Move the Deprecation notes from Getting Started to the top level Chef section. Same Reason.
- Move Chef Manage docs (from 2 different places) into the A1 section. Manage is a premium offering and we sort of called it part of A1 at one point. It's up for debate if that's true or not, but it makes far more sense here than part of concepts and managing the Chef Server.


mjingle (approved):
tenor-233850727

PR #1671, opened 2018-09-24
Improve notes in the DK installation doc

tas50:
Simplify how we suggest installed Xcode
Clarify why / when you need Xcode
Clarify that you don't actually need to be the root user, but a user with admin privs


mjingle (approved):
tenor-194506011

PR #1675, opened 2018-09-25
Link directly to macos / windows DK packages

tas50:
This makes it a tiny bit easier to download.


kagarmoe (approved):

kagarmoe (commented):
Need to get it through buildkite

PR #1676, opened 2018-09-26
.kitchen.yml -> kitchen.yml and remove old drivers

tas50:
Yank the ones that don't exist / work anymore. Update the name of the config file.


mjingle (approved):
tenor-106398955

PR #1677, opened 2018-09-26
Fix 404 from google with a redirect

tas50:
I hope people are off Chef 10 at this point, but just in case now they get redirected to the archived docs.


mjingle (approved):
tenor-219681157

PR #1678, opened 2018-09-26
Shuffle around more things and remove some DK nav items

tas50:
Remove duplicate release notes and deprecation pages that happened with the nav shuffles
Don't name things workstation when we mean DK since that's already getting confusing
Yank more features we've deprecated from the nav bar
Remove a very funky configuration page that had nothing to do with DK


mjingle (approved):
tenor-268015259

PR #1679, opened 2018-09-26
Cleanup + add glossary entries

tas50:
Fix capitalization of tool names
Better describe what DK does. It has nothing to do with developing chef itself
Add cookstyle


mjingle (approved):
tenor-192317054

PR #1682, opened 2018-09-27
Move around the nav menu a bit more

tas50:
Collapse the troubleshooting section that only had 1 sub menu item. Move
run lists into policy since it lines up with environments and all that
fun stuff.


mjingle (approved):
tenor-245379129

PR #1689, opened 2018-10-03
Custom Resource Props and Keywords

sean-horn:
Conflicts between custom resource properties and existing keywords in the resource system, like ":name" result in difficult to debug errors with no indication from the system that the use of ":name" will cause an issue. We'll clarify with a warning here.


mjingle (approved):
tenor-260239540

PR #1694, opened 2018-10-07
Give the user resource the right title

tas50:
This makes it show up better in search and matches our other resources


mjingle (approved):
tenor-264892030

PR #1697, opened 2018-10-08
Massive copyediting for true, false, and Default Values

mjingle:
Signed-off-by: Mary Jinglewski <mjinglewski@chef.io>

### Description

Cleans up "True, False" values to "true, false" and fixes formatting for various **Default Value:**

### Definition of Done

When the dtags are resolved.

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/chef_master/source/debug.rst
+++ b/chef_master/source/debug.rst
@@ -120,7 +120,7 @@ Attributes
 This resource has the following properties:
 
 ``ignore_failure``
-   **Ruby Types:** True, False | **Default Value:** ``false``
+   **Ruby Types:** true, false | **Default Value:** ``false``
 
    Continue running a recipe if a resource fails for any reason.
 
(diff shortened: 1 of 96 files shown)


tas50 (approved):

tas50:
Thanks for fixing them all

mjingle:
Funnily enough, I think I'm 2/3rd of the way through by my estimations. I did find some 'nil' default values, which I kept in the paragraph form. Didn't you want those gone, @tas50? I can keep chipping away at this.

tas50:
We should probably remove nil defaults. They don't really mean much to users

mjingle:
I'll put the nils in another PR. Hopefully that will make reviewing less painful.

PR #1702, opened 2018-10-09
Add timezone resource for Chef 14.6

tas50:
Another resource shipped!


mjingle (approved):
tenor-226553918

PR #1704, opened 2018-10-10
Documenting deprecation of the system_profile plugin

mjingle:
Signed-off-by: Mary Jinglewski <mjinglewski@chef.io>

### Description
Documents upcoming depreciation of system_profile plugin in Ohai

### Definition of Done

When @tas50 says it's okay.

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/chef_master/source/ohai.rst
+++ b/chef_master/source/ohai.rst
@@ -218,6 +218,8 @@ For attributes that contain slashes (``/``) within the attribute value, such as 
 
 Default Plugins
 =====================================================
+.. warning:: The system_profile plugin will be removed from Chef/Ohai 15 in April 2019. This plugin does not correctly return data on modern Mac systems. Additionally the same data is provided by the hardware plugin, which has a format that is simpler to consume. Removing this plugin will reduce Ohai return by ~3 seconds and greatly reduce the size of the node object on the Chef server.
+
 The following list shows the type of plugins that are included with Ohai. See the ``ohai/lib/ohai/plugins`` directory in the version of Ohai installed on your system for the full list:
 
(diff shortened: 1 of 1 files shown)


mjingle:
@tas50 I noticed that system_profile plugin is listed as "system_profiler.rb" in the list of default plugins available. I assume those are the same thing?

tas50 (changes requested):
Instead of this we'll actually want to create an Ohai deprecations number and create a deprecations page for the removal of the plugin. 

Here's a sample removal deprecation:

https://docs.chef.io/deprecations_ohai_ipscopes.html

The deprecation would go here:

https://docs.chef.io/chef_deprecations_client.html

chef_master/source/ohai.rst
```diff
@@ -218,6 +218,8 @@ For attributes that contain slashes (``/``) within the attribute value, such as
 
 Default Plugins
 =====================================================
+.. warning:: The system_profile plugin will be removed from Chef/Ohai 15 in April 2019. This plugin does not correctly return data on modern Mac systems. Additionally the same data is provided by the hardware plugin, which has a format that is simpler to consume. Removing this plugin will reduce Ohai return by ~3 seconds and greatly reduce the size of the node object on the Chef server.
```
tas50, line 221: We usually just create a deprecation page on the deprecations page for these. There's too many to track with warnings.

mjingle: @tas50 Is there a certain code that needs to be used for this on the depreciations page?

PR #1709, opened 2018-10-15
include version introduced

derekgroh:
### Description

includes the documentation on when this resource was introduced following the format of other resources.


mjingle:
@derekgroh Thank you for your contribution! Because this is a PR from a remote fork, our set-up means that we had to include your work to chef-web-docs through other means. Please see https://github.com/chef/chef-web-docs/pull/1723

PR #1716, opened 2018-10-18
Remove Ohai from the list of CLI tools

tas50:
It's not the most important thing in the DK/workstation


mjingle (approved):
tenor-132304171

PR #1721, opened 2018-10-18
Add ChefDK 3.2 and 3.3 release notes

tas50:
I suspect you might want to update the formatting a bit, but this should all work. It only 2 two hours of trial and error to get the silly links to work without blowing up the build.


mjingle (approved):
tenor-223281306

PR #1733, opened 2018-10-26
Update support to include 10.14

tas50:
DK will ship with it next week. Add it now since we know it works.


mjingle (approved):
tenor-145925335

PR #1734, opened 2018-10-29
Move common resource functionality into its own section

tas50:
Split out common properties from the actual resource properties
Also add some missing properties, fix some types, and add some defaults

### DIFF
--- a/chef_master/source/debug.rst
+++ b/chef_master/source/debug.rst
@@ -113,16 +113,11 @@ The log resource has the following actions:
 
 .. end_tag
 
-Attributes
+Properties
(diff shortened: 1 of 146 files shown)


tas50:
@mjingle @kagarmoe Any idea why I get:
```
Warning, treated as error:
/chef-web-docs/chef_master/source/resource_dmg_package.rst:137: (SEVERE/4) Unexpected section title
```

https://github.com/chef/chef-web-docs/blob/e425875526c227caee4bd23f3d8be89979a02533/chef_master/source/resource_dmg_package.rst

I'm rst confuses right now

kagarmoe (approved):

PR #1740, opened 2018-11-02
Clarify A1 custom ssl cert and key behavior

sean-horn:
Automate 1.x does a weird thing where it writes a custom key/cert OVER the existing cert and key contents, if the naming of the cert/key still match.


mjingle (approved):
tenor-68122296

PR #1746, opened 2018-11-07
More resource cleanup from automation

tas50:
Move some more common properties to their own section
Fix some default values
Remove duplicate properties
Fix formatting
Fix wording on the name_property type properties to better explain that they are an optional way to override the resource block's name value. We're still not very good at describing this, but I see a lot of people using them when they don't need them so hopefully this helps.

### DIFF
--- a/chef_master/source/release_notes.rst
+++ b/chef_master/source/release_notes.rst
@@ -2985,6 +2985,13 @@ The launchd resource has the following actions:
 ``:restart``
    Restart a launchd managed daemon or agent.
 
+``:nothing``
+   .. tag resources_common_actions_nothing
+
+   Define this resource block to do nothing until notified by another resource to take action. When this resource is notified, this resource block is either run immediately or it is queued up to be run at the end of the Chef Client run.
+
+   .. end_tag
+
 .. end_tag
 
 Properties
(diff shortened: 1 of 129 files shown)


chef_master/source/release_notes.rst
```diff
@@ -2985,6 +2985,13 @@ The launchd resource has the following actions:
 ``:restart``
    Restart a launchd managed daemon or agent.
 
+``:nothing``
+   .. tag resources_common_actions_nothing
+
+   Define this resource block to do nothing until notified by another resource to take action. When this resource is notified, this resource block is either run immediately or it is queued up to be run at the end of the Chef Client run.
```
kagarmoe: s/until/unless ?

kagarmoe: Consider: This resource block does not act unless notified by another resource to take action. Once notified, this resource block either runs immediately or is queued up to run at the end of the Chef Client run.

kagarmoe (changes requested):

chef_master/source/release_notes.rst
tas50: Updated

kagarmoe (approved):

kagarmoe:
tenor-215320877

PR #1754, opened 2018-11-09
Remove config information for Audit mode

tas50:
This was a beta feature that never came out of beta and is being removed
from the client entirely. No one needs to know about these config
options and it only confuses users since we have a modern "audit" mode
via the audit cookbook, which is an entirely different thing.


mjingle (approved):
tenor-54863836

PR #1752, opened 2018-11-09
More resource docs cleanup

tas50:
- Due to a bug in the windows_task resource wasn't showing up with the resource inspector before. This is fixed so now we get to fix some issues.
- Remove some very old Chef 12 "new in" notes that just pollute the docs at this point
- Fix some formatting
- Add more common options
- Fix the italic "new in" notes

Signed-off-by: Tim Smith <tsmith@chef.io>

PR #1755, opened 2018-11-09
Don't mention very old Chef 12 releases

tas50:
I left in more recent notes about Chef 12, but these are just ancient and we shouldn't litter our notes with information that isn't relevant anymore.


tas50:
Pulling into #1752

PR #1756, opened 2018-11-09
platform support: Add missing version, remove EOL versions, fix format

tas50:
Add some Ubuntu 18.04 for a few products which we ship on 18.04 now
Remove FreeBSD 10 support which went EOL 10/31
Consistently use the .x format when we're saying we support all minor versions


mjingle (approved):
tenor-58070131

PR #1758, opened 2018-11-13
Update license example to be a SPDX compliant license

tas50:
apachev2 doesn't mean anything


mjingle (approved):
tenor-162493625

PR #1759, opened 2018-11-13
Add some missing config options

tas50:
There's more to add, but this is getting us closer. Wording is straight from the config file.


mjingle (approved):
tenor-57441771

PR #1766, opened 2018-11-16
Resource cleanup v7

tas50:
Most of this is the windows_service and service resource which received some cleanup in the chef-client codebase. There's other fixes though including better wording for the default value of a name_property type property.


mjingle (approved):
tenor-97855659

PR #1816, opened 2019-01-26
Update resource_common.rst

dwmarshall:
s/use case us/use case is/


mjingle:
@dwmarshall Sharp eyes! Thank you for catching that typo. Because this is a PR from a remote fork, our set-up means that we had to include your work to chef-web-docs through other means. Please see https://github.com/chef/chef-web-docs/pull/1819

PR #1818, opened 2019-01-28
Append to `/etc/hosts` instead of overwriting it

teknofire:
### Description

The original self registration bootstrap script would completely wipe out the existing `/etc/hosts` file, this is a safer default to append the contents to the file instead.


mjingle (approved):
tenor-238427402

PR #1823, opened 2019-01-30
Updates to the DK an chef-client release notes

tas50:
We had a fix in chef-client that didn't actually ship

DK was missing a bunch of stuff


mjingle (approved):
tenor-72830902

PR #1824, opened 2019-01-31
Add EOL dates for Chef 13, Provisioning, and DK 2

tas50:
We know these will all EOL on April 30, 2019.

### DIFF
--- a/chef_master/source/platforms.rst
+++ b/chef_master/source/platforms.rst
@@ -510,7 +510,7 @@ Versions and Status
      - n/a
    * - Chef Client
      - 13.x
-     - GA
+     - Deprecated
      - April 30, 2019
    * - Chef Client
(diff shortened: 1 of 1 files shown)


krasnow:
@tas50 - We will officially announce EOL for Provisioning on Feb 15th, so the EOL date should be +6 months out on that. I'll update it to say August 31, 2019. Otherwise LGTM.

krasnow:
Oh and then I noticed that the dates weren't in the right columns. And we hadn't market Compliance as EOL even though the date had passed. So I fixed all that too.

tas50:
Thanks @krasnow. Anything else we should do here @mjingle @kagarmoe

mjingle (approved):
tenor-173947447

PR #1827, opened 2019-02-05
chef-client extension can time out during enable

sean-horn:
Because the chef-client install success is critical, it should not depend on azure extension installs/enable, possible failures. Let's recommend that folks use a more reliable provisioning scheme, like preinstalls


stuartpreston:
I'm not against putting up a warning in the docs **but** we should be clear about the primary reason why and where it applies. I have tried to put a couple of paragraphs below to explain (please take and re-word appropriately) as it's a bit different from the positioning of the proposed text.

When the Chef VM Extension is provisioned as part of a scale operation for a **VM Scale Set** (VMSS) the Microsoft VM agent on the machine **executes all installations in parallel**[1]. As these extensions can call 'locking' installation mechanisms such as Windows Installer (via installation of an MSI), or Apt updates, a timeout can occur if the total time for all of these activities after waiting for retries etc. exceeds 5 minutes. This limit is set by the ‘enable’ phase of the Microsoft VM agent installed on each machine in Azure and is not configurable.

Deployment of a standard VM (i.e. a VM not in a VM Scale Set) where the extension installs can be sequenced through use of ```dependsOn``` within the ARM template language[2]) are much less likely to show issues with extension install. For standard VMs the installation success rate may be improved by adjusting the environmental conditions such as using larger machine sizes, faster disks and, in some cases, using alternative regions for the deployment to achieve faster download rates from the CDN used to distribute the installation files. For the VMSS the non-deterministic nature of the extension sequencing is the most influencing factor for success rate. So for VMSS-based scenarios we recommend pre-installing Chef Client onto the VM and/or using an image-based deployment process (using tools such as Packer[3][4]) to create the base VM disk image.

References:
[1] https://msftstack.wordpress.com/2016/05/12/extension-sequencing-in-azure-vm-scale-sets/
[2] https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-define-dependencies 
[3] https://docs.microsoft.com/en-us/azure/virtual-machines/linux/build-image-with-packer
[4] https://docs.microsoft.com/en-us/azure/virtual-machines/windows/build-image-with-packer

IanMadd:
In the spirit of getting this PR merged someday, does this alternative warning text make sense and adequately address @stuartpreston's concerns? Thoughts? @sean-horn 

> When the Chef VM extension is provisioned as part of a scale operation for a VM Scale Set, we suggest not using the Chef Infra Client extension and instead pre-install Chef Infra Client onto the VM and/or use an image-based deployment process, such as [Packer](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/build-image-with-packer), to create the base VM disk image.
>
> The Microsoft VM agent on the machine executes all installations in parallel and these extensions can call 'locking' installation mechanisms such as Windows Installer (via installation of an MSI) or Apt updates. A timeout can occur if the total time for all of these activities exceeds 5 minutes and in some regions that is not enough time to get the Chef Infra Client package and enable the Chef Infra Client extension.

stuartpreston:
These look like accurate statements to me, thanks @IanMadd

IanMadd (approved):
tenor-4466293

PR #1832, opened 2019-02-11
Remove Juniper marketing copy from Chef for Junos OS docs

pmocek:
### Description

Neither what Juniper Networks do (which is to sell hardware and software, not "solutions" to unspecified problems) nor anyone's assessment of their lead in doing so among competing equipment sellers is relevant to the description of Chef for Junos OS.  Link to the Wikipedia entry about the company upon first mention and remove the irrelevant marketing-speak.


kagarmoe:
We cannot explain how profoundly we agree with this PR, but we have an internal blocking step that we need to finish before we can de-cruft. We'll remove the Juniper docs once we've resolved the rest.

tenor-31899726

PR #1842, opened 2019-03-06
Remove so old / bad advice on node.set

tas50:
This was Chef 11 era advice and what it suggested will actually break Chef 14 users.

### DIFF
--- a/chef_master/source/ruby.rst
+++ b/chef_master/source/ruby.rst
@@ -612,7 +612,7 @@ Constructs to Avoid
 -----------------------------------------------------
 Avoid the following patterns:
 
-* ``node.normal` - Avoid using attributes at normal precedence since they are set directly on the node object itself, rather than implied (computed) at runtime.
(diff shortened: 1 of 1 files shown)


chef_master/source/ruby.rst
```diff
@@ -612,10 +612,8 @@ Constructs to Avoid
 -----------------------------------------------------
 Avoid the following patterns:
 
-* ``node.set`` / ``normal_attributes`` - Avoid using attributes at normal precedence since they are set directly on the node object itself, rather than implied (computed) at runtime.
-* ``node.set_unless`` - Can lead to weird behavior if the node object had something set. Avoid unless altogether necessary (one example where it's necessary is in ``node['postgresql']['server']['password']``)
+* ``node.normal` - Avoid using attributes at normal precedence since they are set directly on the node object itself, rather than implied (computed) at runtime.
```
mjingle:
```suggestion
* ``node.normal`` - Avoid using attributes at normal precedence since they are set directly on the node object itself, rather than implied (computed) at runtime.
```

mjingle (approved):
tenor-137127389

PR #1846, opened 2019-03-08
Fix of pessimistic locking example for chef_version

artem-sidorenko:
minor version is required too

### DIFF
--- a/chef_master/source/config_rb_metadata.rst
+++ b/chef_master/source/config_rb_metadata.rst
@@ -68,7 +68,7 @@ This configuration file has the following settings:
 
    .. code-block:: ruby
 
-      chef_version '~> 14'
+      chef_version '~> 14.0'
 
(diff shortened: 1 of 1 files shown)


kagarmoe (approved):

artem-sidorenko:
somehow this example is right and it works exactly this way: https://travis-ci.org/dev-sec/chef-os-hardening/jobs/503798873

```
FATAL: Chef::Exceptions::CookbookChefVersionMismatch: Cookbook 'os-hardening' version '3.2.1' depends on chef version ["~> 14"], but the running chef version is 15.0.178
```
Not clear why...

mjingle:
@artem-sidorenko Thank you for catching that issue with the example used, and letting us know about it! Because this is a PR from a remote fork, our set-up means that we had to include your work to chef-web-docs through other means. Please see https://github.com/chef/chef-web-docs/pull/1869.

PR #1848, opened 2019-03-12
Deprecating LC_ALL from locale resource

Nimesh-Msys:
Signed-off-by: Nimesh-Msys <nimesh.patni@msystechnologies.com>

### Description

Added a deprecation doc to inform users that we no longer support changing *lc_all* locale. Instead we may change other specific *lc_* property as needed

### Issues Resolved

Based on the [following conversations](https://github.com/chef/chef/pull/8100), it would be a possible solution for [7904](https://github.com/chef/chef/issues/7904)

### Output

chef_deprecations_client

deprecations_lc_all

### Check List

- [x] Spell Check
- [x] Local build
- [x] Examine the local build
- [x] All tests pass


chef_master/source/chef_deprecations_client.rst
```diff
@@ -163,6 +163,10 @@ All Deprecations
     - Resource(s) in a cookbook collide with the same resource(s) now included in Chef.
     - XX.X
     - 15.0
+  * - `CHEF-27 </deprecations_locale_lc_all.html>`__
+    - Deprecation of lc_all from locale resource
+    - 15.0
+    - 15.1
```
btm: Let's plan to deprecate this in 16.0.

chef_master/source/deprecations_locale_lc_all.rst
```diff
@@ -0,0 +1,47 @@
...
+=======================================================================
+`[edit on GitHub] <https://github.com/chef/chef-web-docs/blob/master/chef_master/source/deprecations_locale_lc_all.rst>`__
+
+.. tag deprecations_locale_lc_all
+
+Setting *LC_ALL* variable is NOT recommended and can break Chef's parsing of command output in unexpected ways. Use one of the more specific *LC_* properties as needed.
+
+.. end_tag
+
+This deprecation warning was added after Chef 15.0.185. Support for property `lc_all` will likely be removed by Chef 15.1.0.
```
btm: I don't think we need this line, I think this information is generated in the docs from lines in `chef_master/source/chef_deprecations_client.rst`.

chef_master/source/deprecations_locale_lc_all.rst
```diff
@@ -0,0 +1,47 @@
+=======================================================================
+Deprecation: Deprecation of lc_all from locale resource (CHEF-27)
+=======================================================================
+`[edit on GitHub] <https://github.com/chef/chef-web-docs/blob/master/chef_master/source/deprecations_locale_lc_all.rst>`__
+
+.. tag deprecations_locale_lc_all
+
+Setting *LC_ALL* variable is NOT recommended and can break Chef's parsing of command output in unexpected ways. Use one of the more specific *LC_* properties as needed.
```
btm: Setting the *LC_ALL* variable is NOT recommended. As a system-wide setting, *LANG* should provide the desired behavior. *LC_ALL* is intended to be used for temporarily troubleshooting issues rather than an every day system setting. Changing *LC_ALL* can break Chef's parsing of command output in unexpected ways. Use one of the more specific *LC_* properties as needed.

chef_master/source/deprecations_locale_lc_all.rst
```diff
@@ -0,0 +1,47 @@
...
+Remediation
+=======================================================================
+
+Though it is inadvisable, still if there is a need to set *LC_ALL*, you may set this environment variable manually into any one of the files listed below.
+
+.. code-block:: ruby
+
+	export LC_ALL="<locale_name>"
+
+1. /etc/default/locale
```
btm, line 25: I would add this:

> Setting *LC_* variables varies by platform, but common locations to configure *LC_* variables are:

And then drop the note on "priority order" below.

chef_master/source/deprecations_locale_lc_all.rst
```diff
@@ -0,0 +1,47 @@
...
+
+
+To check the locale value, run
+
+.. code-block:: ruby
+
+	locale -v
+
+.. warning::
+	- Changes on any other file might only be applicable till current session of shell.
```
btm: drop this warning. this is probably a step too far in trying to teach unix than provide a workaround for the deprecation of the `lc_all` property.

chef_master/source/chef_deprecations_client.rst
btm: for the deprecation version let's set to 14.12, and we will backport the deprecation into the Chef 14 release we will make in April/May. And set the second number to 16.0 as I commented/

chef_master/source/deprecations_locale_lc_all.rst
btm: Apparently the other file doesn't do anything automated for us, so we need to keep this line. But lets say:

> This deprecation warning was added ln Chef 14.12. Support for property `lc_all` will be removed for Chef 16.0

Nimesh-Msys: Its not automated. Have added the comment manually.

chef_master/source/deprecations_locale_lc_all.rst
Nimesh-Msys: Updated

chef_master/source/deprecations_locale_lc_all.rst
Nimesh-Msys: True. Updated

Nimesh-Msys:
@btm , I've updated this PR as per the review comments. Tested and updated the output in Description.
Also squashed the changes into single commit. Could you please take a look and let me know in case of any issues.

mjingle:
@Nimesh-Msys Thank you for your documentation contribution and the level of detail in your PR! Because this is a PR from a remote fork, our set-up means that we had to include your work to chef-web-docs through other means. Please see: https://github.com/chef/chef-web-docs/pull/1858

PR #1854, opened 2019-03-20
load_current_value examples require block variable

pezius:
### Description

Change updates the examples of `load_current_resource` to function (they do not as written).  They require the explicit definition of a block variable to work correctly.

### Definition of Done
n/a

### Issues Resolved

n/a

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/chef_master/source/custom_resources.rst
+++ b/chef_master/source/custom_resources.rst
@@ -546,9 +546,9 @@ For example, a custom resource defines two properties (``content`` and ``path``)
    property :content, String
    property :path, String, name_property: true
 
-   load_current_value do
-     if ::File.exist?(path)
-       content IO.read(path)
+   load_current_value do |desired|
+     if ::File.exist?(desired.path)
+       content IO.read(desired.path)
      end
    end
 
@@ -581,10 +581,10 @@ The ``converge_if_changed`` method may be used multiple times. The following exa
    property :content, String
    property :mode, String
 
-   load_current_value do
-     if ::File.exist?(new_resource.path)
-       content IO.read(new_resource.path)
-       mode ::File.stat(new_resource.path).mode
+   load_current_value do |desired|
+     if ::File.exist?(desired.path)
+       content IO.read(desired.path)
+       mode ::File.stat(desired.path).mode
      end
    end
 


chef_master/source/custom_resources.rst
```diff
@@ -581,10 +581,10 @@ The ``converge_if_changed`` method may be used multiple times. The following exa
    property :content, String
    property :mode, String
 
-   load_current_value do
-     if ::File.exist?(new_resource.path)
-       content IO.read(new_resource.path)
-       mode ::File.stat(new_resource.path).mode
+   load_current_value do |desired|
```
MarkGibbons, line 584: Why not use new_resource as the block variable name?  OK, I see you were following the example given in the Block Arguments doc.  The block arguments doc is also very unclear.  "Use a block argument that contains all of the properties of the requested resource." Really should read "Use a block argument, it will contain all of the properties of the requested resource".  I think the block arguments example needs to be updated also.

chef_master/source/custom_resources.rst
```diff
@@ -546,9 +546,9 @@ For example, a custom resource defines two properties (``content`` and ``path``)
    property :content, String
    property :path, String, name_property: true
```
MarkGibbons, line 547: Would this work if path had the identify: true option and no block argument was used?

MarkGibbons: That would show the other method I think works for this.

chef_master/source/custom_resources.rst
MarkGibbons: I ran into this situation helping some people write custom resources this week.  We had to read the resource code to figure out how load_current_value was implemented to get it to work.

chef_master/source/custom_resources.rst
MarkGibbons: I tested these examples.  This example does work, but only because path has name_property: true specified.  That name_property exposes the variable in the load_current_value block is not documented, so this working is mostly luck.

MarkGibbons:
The changes here are part of #2118.

tas50:
Closing this in favor of #2118

PR #1862, opened 2019-04-01
locale resource: Support LC_ Env Vars

Nimesh-Msys:
Signed-off-by: Nimesh-Msys <nimesh.patni@msystechnologies.com>

### Description

Doc changes related to [8324](https://github.com/chef/chef/pull/8324).

*lc_all* property has been deprecated from locale resource, instead we have provided support to set values for other LC env variables. For that, we have introduced *lc_env* property, of Hash type that could accept any LC var as the key and language as its value.

### Issues Resolved

Based on the [following conversations](https://github.com/chef/chef/pull/8100), it would be a possible solution for [7904](https://github.com/chef/chef/issues/7904)
This would also complete the [deprecation](https://github.com/chef/chef-web-docs/pull/1848) [docs](https://github.com/chef/chef-web-docs/pull/1858)

### Check List

- [x] Spell Check
- [x] Local build
- [x] Examine the local build
- [x] All tests pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ docker-preview: docker-build
 	docker run -it -v $(shell pwd):/chef-web-docs \
 		-w /chef-web-docs/public \
 		-p 8000:8000 chefes/buildkite \
-		bash -c 'export PATH=$$PATH:/chef-web-docs/doctools; python -m SimpleHTTPServer'
+		bash -c 'export PATH=$$PATH:/chef-web-docs/doctools; python -m http.server'
 
 docker-dtags:
 	docker run -it -v $(shell pwd):/chef-web-docs \
--- a/chef_master/source/resource_locale.rst
+++ b/chef_master/source/resource_locale.rst
@@ -14,8 +14,8 @@ The locale resource has the following syntax:
 .. code-block:: ruby
 
   locale 'name' do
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ docker-preview: docker-build
 	docker run -it -v $(shell pwd):/chef-web-docs \
 		-w /chef-web-docs/public \
 		-p 8000:8000 chefes/buildkite \
-		bash -c 'export PATH=$$PATH:/chef-web-docs/doctools; python -m SimpleHTTPServer'
+		bash -c 'export PATH=$$PATH:/chef-web-docs/doctools; python -m http.server'
```
btm, line 25: what was this change for? is this a consequence of a python / library version change?

Nimesh-Msys: Yes, `python -m SimpleHTTPServer` runs well until Python-3. But in docker container `Python 3.7.0` was getting installed which `uses http.server module`. Therefore, when I changed the command to `python -m http.server`, I was able to see the output.

mjingle:
@Nimesh-Msys Thank you for your documentation contribution and finishing up these set of docs! Because this is a PR from a remote fork, our set-up means that we had to include your work to chef-web-docs through other means. Please see: https://github.com/chef/chef-web-docs/pull/1866

PR #1872, opened 2019-04-15
policy_group and policy_name description around the wrong way

atward:
### Description

Swap text descriptions of `policy_group` and `policy_name`


atward:
Found in multiple locations, including `release_notes.rst`, from [12.5](https://docs.chef.io/release_notes.html#specify-policy-revision). Will leave it to mods to determine if release_notes should retrospectively edited.

tas50 (approved):

mjingle:
@atward Thank you for catching those switched description fields and letting us know about it! Because this is a PR from a remote fork, our set-up means that we had to include your work to chef-web-docs through other means. Please see https://github.com/chef/chef-web-docs/pull/1878.

PR #1895, opened 2019-05-06
Chef needs a cookbooks directory, etc.

RichMorin:
### Description

It adds instructions on how to prepare for the cookbook creation and use.

### Definition of Done

?

### Issues Resolved

The page didn't say anything about changing the PATH, creating a cookbooks directory, et.

### DIFF
--- a/chef_master/source/quick_start.rst
+++ b/chef_master/source/quick_start.rst
@@ -7,15 +7,23 @@ For the quickest way to get started using Chef:
 
 #. Install the Chef development kit: https://downloads.chef.io/chefdk/.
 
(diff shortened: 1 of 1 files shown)


tas50:
@RichMorin Thanks for the the PR, but we're not going to merge this particular one. The goal of the getting started page is the provide a quick getting started example for how a user can create and execute a single cookbook. We're very purposefully trying to avoid getting into the mono-repo vs. single cookbook in a repo workflow debates within this page. It is however entirely valid to just have a single cookbook in a repo, and that's how many users operate.

PR #1896, opened 2019-05-07
Update resources using automation

tas50:
Hand syncing from the auto generated content using Chef 15.0


mjingle (approved):
tenor-42049364

PR #1906, opened 2019-05-14
Fix instructions for chef infra acceptance

mivok:
The original instructions said to put an equals sign in your client.rb file, but this doesn't work. However, putting `chef_license 'accept'` without an equals sign does work, and also matches the format for the rest of the configuration.

### DIFF
--- a/chef_master/source/chef_license_accept.rst
+++ b/chef_master/source/chef_license_accept.rst
@@ -71,7 +71,7 @@ The same license applies to all products, but each product must have its own lic
 
 Chef Infra Client
 -----------------------------------------------------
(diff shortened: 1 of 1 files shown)


tyler-ball (approved):
Yup, totally my bad - thanks for catching and fixing this!

mjingle:
@mivok You'll need to add a DCO to your commit, but thank you for your PR! See: https://blog.chef.io/2016/09/19/introducing-developer-certificate-of-origin/

mivok:
Oops, that's what I get for using the web interface to make the change. One moment.

PR #1934, opened 2019-05-30
Remove Chef 13 and ChefDK 2 as they are now EOL

btm:
Some distributions (those without word marks held by Chef Software) do not require Chef EULA
acceptance directly, but do require acceptance for other distributions
for full functionality.

### DIFF
--- a/chef_master/source/versions.rst
+++ b/chef_master/source/versions.rst
@@ -73,25 +73,27 @@ Use of these and later versions of these distributions must be in accordance wit
      - GA
      - n/a
    * - Chef Workstation
-     - 0.3
+     - 0.4
      - Releasing Soon
      - n/a
    * - ChefDK
      - 4.x
-     - Releasing Soon
+     - GA
      - n/a
-   * - Supermarket
+   * - Supermarket¹
      - 4.x
      - Releasing Soon
      - n/a
-   * - Chef Backend
+   * - Chef Backend¹
      - 3.x
      - Releasing Soon
      - n/a
-   * - Push Jobs
+   * - Push Jobs¹
      - 3.x
      - Releasing Soon
      - n/a
+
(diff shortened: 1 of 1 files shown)


mjingle:
@btm Would the Chef Client and Chef DK that you deleted  be better re-labeled as "EOL" in the table instead?

mjingle:
@btm Alternatively, we could also put the deleted information into the End of Life (EOL) Products table at the bottom of the page.

btm:
@mjingle I originally moved those down to EOL, but then I saw this:

https://docs.chef.io/versions.html#versions-and-status
> Unless otherwise stated, versions older than those listed below are EOL.

So I removed htem.

mjingle:
@btm Understood. By removing them completely from the chart, there is no reference for the two products elsewhere (beyond Supported Free Distributions and Supported Commercial Distributions). Also, the Deprecated Products and Versions section has: "The following products are deprecated or end of life." Would it make sense then, instead, to change the status from "Deprecated" to "EOL" in the Lifecycle Status column of the Deprecated Products and Versions table? I'm hoping to maintain a frame of reference.

This could also be solved by updating the product's version with its new deprecation date? e.g. ChefDK 3.x | EOL Date: TBD

btm:
@mjingle it feels odd to me to leave EOL products in the "Deprecated Products and Versions" section when there is a separate "End of Life (EOL) Products" section.

I think my original gut reaction to put something like this in the EOL section is the most clear?
|  Chef Client | 13 and under | EOL | April 30, 2019 |

mjingle:
@btm I would go with that gut reaction! Products are listed alphabetically in both tables.

btm:
@mjingle force pushed and rebased.

mjingle (approved):
tenor-82385417

PR #1932, opened 2019-05-30
Remove support for Ubuntu 14.04

tas50:
Ubuntu 14.04 went EOL. Note that this remove support for PPC Ubuntu builds since they were only supported on Ubuntu 14.04.

PR #1935, opened 2019-05-30
Drop Ubuntu 14.04

btm:
Ubuntu went EOL on 4/30/19, and per our policy that makes it an EOL
platform for us.

This also marks the end of ppc64le builds for Chef Server.


btm:
partially duplicates #1932, but also adds 18.04 and removes x86 for Windows.

mjingle (approved):
tenor-213318947

PR #1937, opened 2019-05-30
add example to sysctl

echohack:
Signed-off-by: echohack <echohack@users.noreply.github.com>


lamont-granquist:
There should probably be an example here that all these work equivalently (at least I think they do?):

```
sysctl 'net.ipv4.ip_local_port_range' do
  value '9000 65500'
end

sysctl 'net.ipv4.ip_local_port_range' do
  value [ '9000', '65500' ]
end

sysctl 'net.ipv4.ip_local_port_range' do
  value [ 9000, 65500 ]
end
```

but that may be its own segment.  i think that'd be more useful than the swappiness example.

YMMV.

echohack:
@lamont-granquist Can we please merge this and then address the extra examples in another PR? Otherwise this is gonna rot on the vine.

mjingle (approved):
tenor-137339502

PR #1943, opened 2019-06-04
Refactors docs for `ctl chef client options format`

kagarmoe:
### Description

This change will be more controversial, as all of the affected pages have high traffic.

However, this is my hill: CTL/CLI documentation should appear **only** on the CTL/CLI reference pages.

Why? Because duplicating content endlessly is one of the major problems for chef-web-docs. If we serve CTL/CLI docs **only** on their corresponding reference pages then 1) Users will know where to go to find the reference docs 2) search will be more effective and 3) one day I will automate these docs.

### Issues Resolved

### Check List

- [ ] Spell Check
- [x] Local build
- [x] Examine the local build
- [ ] All tests pass


mjingle (approved):
tenor-29519086

PR #1953, opened 2019-06-05
Fix description of :create action

karlamrhein:
Fixed description of :create action.  It was the same as :create_if_missing.

Obvious fix (invoking the "obvious fix" rule)

### DIFF
--- a/chef_master/source/resource_cron_d.rst
+++ b/chef_master/source/resource_cron_d.rst
@@ -70,7 +70,7 @@ Actions
 The cron_d resource has the following actions:
 
 ``:create``
-   Default. "Add a cron definition file to /etc/cron.d, but do not update an existing file.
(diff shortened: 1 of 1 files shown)


mjingle:
@karlamrhein Thank you for spotting this and letting us know about it! This is indeed a mis-paste from our code. (The `"` kind of gave it away, eh?) The correct description for this action can be found here: https://github.com/chef/chef/blob/master/lib/chef/resource/cron_d.rb#L170
Your PR should be updated accordingly to match the description in the code. I can do that if you would like.

karlamrhein:
yes, the ```"``` was a clue!  

i just updated the PR to match the code.

mjingle:
@karlamrhein Fantastic. Thank you so much for your contribution!

mjingle (approved):
tenor-199182392

PR #1964, opened 2019-06-09
Correction to default for `sensitive` property

joeselway:
Contradicts line 23. Verified `true` as correct default per source https://github.com/chef/chef/blob/master/lib/chef/resource/windows_ad_join.rb line 62.

Obvious fix.


mjingle:
@joeselway Thank you for this PR and for your patience! Originally, this fix affected a number of other files due to our current dtag structure. I had to investigate the best way to address this first. We appreciate you letting us know about this!

mjingle (approved):
tenor-23645957

PR #1981, opened 2019-06-18
Kg/resource common actions nothing

kagarmoe:
### Description

Fixes nesting of dtag `resourse_common_actions_nothing` in most cases by removing containing tags. 
- Removed all containing tags from `provisioning.rst`, which turned many tags into single-use.
- Removed single-use tags
- s/Chef provisioning/Chef Provisioning
- Changed `tag note_provisioning` to `tag EOL_provisioning`

Nested dtags for this tag still exist in:
- `tag resource_breakpoint_actions`
- `tag resource_log_actions`
- `tag resource_ohai_actions` 

Two of these uses are located on `debug.rst`. However, the debug page is ranked #53 for 12-month use and requires further examination of user behavior before refactoring.

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/chef_master/source/ctl_chef.rst
+++ b/chef_master/source/ctl_chef.rst
@@ -711,16 +711,24 @@ None.
 
 chef provision
 =====================================================
+.. tag EOL_provisioning
+
+This functionality was available with Chef Provisioning and was packaged in the Chef development kit.
+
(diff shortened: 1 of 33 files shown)


IanMadd (approved):
tenor-174733569

PR #1994, opened 2019-06-25
Add notice about template cookbook locations

emachnic:
When developing a custom resource, I found that the cookbook in which I was trying to call the resource could not find the templates and realized that I needed to specify the cookbook location where those templates live.

### Description

Adds a note about specifying the cookbook location in a custom resource template

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


tas50 (approved):

tas50:
Thanks for adding the comment. I think we may shuffle this all around in the near future, but this is great for today and it'll remind us in the future to fully document how templates behave in a resource world.

tas50:
I created a cookstyle issue for this since I think we can auto detect it: https://github.com/chef/cookstyle/issues/160

IanMadd (approved):
Thanks for the contribution @emachnic!

tenor-202987062

PR #1995, opened 2019-06-25
User endpoint

MarkGibbons:
### Description

Update the /users api endpoint documentation.

    *  The /users GET method accepts parameters, add them to the documentation.
    *  The output returned from creating a user has a different format.
    *  The output returned from getting a user has many more fields and name was renamed to username at some point.
    *  /users/NAME POST to create a user is not a thing. The api always returns a 405.

    The /user/NAME PUT documentation is still wrong, haven't figured that out yet. 

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


markan:
Thanks for this; it looks like a real improvement! I want to do one last read through the erlang code before final signoff. I'll try and get to that later today or first thing tomorrow.

chef_master/source/api_chef_server.rst
```diff
@@ -634,7 +632,18 @@ GET
...
+.. list-table::
+   :widths: 200 300
+   :header-rows: 1
+
+   * - Parameter
+     - Description
+   * - ``email=jane@chef.com``
+     - Filter the users returned based on their email id.
+   * - ``external_authentication_uid=jane@chef.com``
+     - Filter the users returned based on their external login id.
```
markan, line 646: Note: if the ``verbose`` parameter is passed it returns a user list with "email", "first_name", "last_name" fields. If this flag is set the email and external_authentication_uid parameters are ignored.

chef_master/source/api_chef_server.rst
```diff
@@ -712,8 +721,9 @@ with a request body similar to:
 
 where:
 
-* ``name`` must begin with a lower-case letter or digit, may only contain lower-case letters, digits, hyphens, and underscores. For example: ``chef``.
-* ``email``, ``name``, and ``password`` are all required to be present and have a value.
+* ``username`` must begin with a lower-case letter or digit, may only contain lower-case letters, digits, hyphens, and underscores. For example: ``chef``.
+* ``display_name``, ``email`` and  ``username`` are required to be present and have a value.
```
markan: Some additional validation is done on the email to attempt to restrict it a bit. The somewhat gruesome regex is here: https://github.com/chef/chef-server/blob/master/src/oc_erchef/apps/chef_objects/src/chef_user.erl#L224. 
Usernames are a dotted list of ```a-z0-9!#$%&'*+/=?^_`{|}~-``` elements, while the host portion can be a dotted list of '[a-z0-9](?:[a-z0-9-]*[a-z0-9])' elements, or an IPv4 address, or some other weird stuff.

chef_master/source/api_chef_server.rst
markan: Text like the following would work:
```
* - ``verbose=true``
   - Returns a user list with "email", "first_name", "last_name"
      fields. If this flag is set the email and
      external_authentication_uid parameters are ignored.
```

chef_master/source/api_chef_server.rst
markan: Possible text:
* ``email`` is required to be present and fit a loose definition of a valid email address. The basic form is a username@host structure, where username is a dot separated list of ```a-z0-9!#$%&'*+/=?^_`{|}~-``` elements, while the host name can be a dot separated list of ```[a-z0-9](?:[a-z0-9-]*[a-z0-9])``` elements, or an IPv4 address, or a few other things.
I'm skeptical this actually is correct in a unicode world.

markan (changes requested):
First of all, thanks for this PR. Sorry for the delay, but I wanted to read the actual code before signing off on it. I added a few minor tweaks with suggested text based on that.

Two TODOs that we should look into:
1) It's clear that we've let our RAML docs (https://github.com/chef/chef-server/tree/master/src/oc_erchef/raml-docs) atrophy; we should review and reconcile them with this doc. 
2) The regex for the email seems sketchy, we should review it again to make sure it makes sense in a modern era.
Neither of these should be construed as blocking this PR; instead a note for me to file issues.

chef_master/source/api_chef_server.rst
MarkGibbons: I thought about including a link to the code to document the email regex, but that seems like a maintenance nightmare.  Instead I included a simple description of the requirements.

chef_master/source/api_chef_server.rst
MarkGibbons: Updated.

MarkGibbons: I'm still checking the format of the verbose response.

mjingle:
@MarkGibbons It's looking good so far! Can you please include a DCO in your commits? There's more information on why we require the DCO here: https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

MarkGibbons:
Added DCO, Thanks

chef_master/source/api_chef_server.rst
markan: That seems entirely reasonable to me.

chef_master/source/api_chef_server.rst
mjingle: Is this good to go @MarkGibbons?

markan (approved):
I'm +1 to merging this. Thanks again @MarkGibbons for taking this on; this kind of work is hugely valuable.

mjingle:
Let me resolve this doc build issue and we'll get this merged! Thank you so much for your contribution @MarkGibbons!

PR #2002, opened 2019-06-27
Remove docs for the EOL Audit Mode

tas50:
We introduced this as an experimental feature in 2015. It has to be turned on via the config. Shortly after that we released InSpec and this feature was never worked on after that or really used by anyone. We removed it in DK 4 / Chef 15. We should entirely wipe it from our memories.


IanMadd (approved):
Thanks!

PR #2003, opened 2019-06-27
Remove some older / broken knife plugins from our list

tas50:
Remove the Chef written plugins that we don't maintain anymore
Remove knife-voxel as the company is no more
Remove knife-community as knife supermarket is built into chef now
Remove knife-baremetalcloud as it hasn't been updated in 7 years


IanMadd (approved):
tenor-156681197

PR #2005, opened 2019-06-28
Update the list of knife cloud plugins

tas50:
This is a better list of modern plugins


mjingle (approved):
tenor-244283251

PR #2006, opened 2019-06-28
Remove the nav bar element for knife spork.

tas50:
The knife-spork plugin ships with Chef, but as we move to model where we encourage users to adopt Policyfiles and move to our modern workflow we should not be introducing new users to spork. This is another step to limiting the mental scope of our kit with new users.


tas50:
Tagging @krasnow to discuss this change

kagarmoe (approved):

krasnow:
👍 This kind of cleanup is great. Thanks Tim.

PR #2008, opened 2019-06-28
Remove the Chef Client and Knife plugins

tas50:
Chef Client doesn't actually have plugins. This was just an odd list of handlers and some random tools that don't make sense in 2019.

The knife plugins in this list are of real limited value. Most of these are not maintained and I would guess many no longer work. We really shouldn’t list these here since we have this on the Supermarket now.


tas50:
This is another one that probably needs @krasnow to sign off on. This removes a lot of community content that has not been very well maintained and doesn't provide a lot of value to new users. We share some of this on the supermarket and we shouldn't duplicate it here.

krasnow:
👍 Approved. Let's clean this up. Thanks y'all.

mjingle (approved):
tenor-225984954

PR #2016, opened 2019-07-09
Add ohai plugin directory to cookbooks, clean up a little

btm:
Chef cookbooks may contain custom ohai plugins in the 'ohai' directory which
will be installed and loaded as part of a client run.

Also, cleaned up some confusion and duplication. There's still more to
do but this is better.


tas50 (approved):
Has some test failures, but looks like a solid improvement

PR #2032, opened 2019-07-17
Replace .kitchen.yml with kitchen.yml

tas50:
We still had a few left over.


mjingle (commented):
tenor-21639308

mjingle (approved):
tenor-21639308

PR #2038, opened 2019-07-24
Move to a Habitat/Hugo based site deployment

schisamo:
Now that we are moving docs.chef.io to Hugo we can take advantage of our
`cd_hab_hugo_static_site` Terraform module which was developed for
deploying automate.chef.io. This approach executes the underlying Hugo
build using Habitat (which produces a versioned artifact). We can also
take advantage of Habitat channels for promotions!


terraform/Makefile
```diff
@@ -1,19 +1,54 @@
...
+
+remote-state: init
 	# We make sure that existing local state is deleted before configuring
 	# the remote state. If it is not removed state gets contaminated between
 	# environments.
-	cd $(TF_DIRECTORY) && \
+
 	rm -rf .terraform && \
-	terraform init -backend-config="key=chef-web-docs/$(TF_STATE)"
+	terraform init -backend-config="key=a2-automate-chef-io/$(TF_STATE)"
```
tduffield: still referencing automate

tduffield (changes requested):

terraform/Makefile
schisamo: fixed!

tduffield (approved):

PR #2047, opened 2019-08-01
Update pages to work with Swiftype

IanMadd:

### Description

This does two things:

1. Deletes all pages listed in the robots.txt file
Robots, including Swiftype, will not crawl through files listed in the robots.txt. But, if a page was crawled and indexed before it was added to the robots.txt file, then it won't be updated in the index even if it's deleted.

2. Adds a `noindex` metatag to EOL and deprecated pages. 
This will tell Swiftype to delete its indexes of these pages and it won't re-index those pages. Other robots should also avoid indexing those pages as well. 

### Check List

- [ ] Spell Check
- [X] Local build
- [X] Examine the local build
- [ ] All tests pass


mjingle (approved):
tenor-187296860

PR #2048, opened 2019-08-05
add noindex meta tags to several more files

IanMadd:

### Description

Add `robots: noindex` tags to several more files that I missed in #2047 

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


kagarmoe:
We should keep the audit cookbook discoverable, as well as the two resources. `resource_deploy` gets a fair amount of traffic (101 pageviews last week). `resource_erlang_call` less so (1), but as long as we support it, we should make it easier to find.

In the past, docs "hid" things from users with the robots.txt or by dropping from the left nav. This is a bad practice, as we lose the context for these decisions over time.

Possibly, we could add a metadata field to the header to explain decisions to `no-index` active docs in the future and keep the history of decisions on the documents?

kagarmoe (approved):
tenor-63869973

PR #2049, opened 2019-08-06
Update many of our supported platforms

tas50:
Our downloads page and this page did not match


miah (approved):
Thanks @tas50

PR #2054, opened 2019-08-08
More resource cleanup to match reality

tas50:
Fix spacing
Remove some bogus properties
Fix some package resources that don't support multipackage so we can't tell folks to pass in arrays
Add the :nothing action where it was missing
Redo a few of the resource pages that had never received the new format


lamont-granquist (approved):

IanMadd (approved):
tenor-238025682

PR #2055, opened 2019-08-09
More updates to our supported platforms

tas50:
Chef Infra doesn't support Ubuntu i386 anymore now that 14.04 is gone
Chef Infra supports RHEL 8 now
DK supports SLES 15
DK does not support Windows 7 / 8.1
InSpec does not support WIndows 7 / 8.1 / 2008 R2
DK works on openSUSE 42 so consider that community supported


IanMadd (approved):
tenor-78389623

PR #2059, opened 2019-08-09
Remove cert advice only important if you're on Chef 11

tas50:
This is letting folks know about new behavior in Chef 12.  We don't need to waste time with this anymore.


IanMadd (approved):
tenor-130167534

PR #2068, opened 2019-08-13
Warn about client_key config.rb relative path spec

sean-horn:

### Description

knife/chef commands using config.rb in most cases may not be aware of what the shorthand `~` should mean. This can result in failure to read the file like this

```
2019-08-13T18:13:36+00:00] WARN: Failed to read the private key ~/.chef/jenkins.pem: #<Errno::ENOENT: No such file or directory @ rb_sysopen - ~/.chef/jenkins.pem>

Error: Failed to upload policy to policy group test

Reason: (Chef::Exceptions::PrivateKeyMissing) I cannot read ~/.chef/jenkins.pem, which you told me to use to sign requests!
```

### Issues Resolved

Let's instead change the docs to warn explicitly that the path to the client_key file be specified as an absolute path from the root.

### Check List

- [X] Spell Check

### DIFF
--- a/chef_master/source/config_rb.rst
+++ b/chef_master/source/config_rb.rst
@@ -49,7 +49,7 @@ This configuration file has the following settings:
    A directory that contains additional configuration scripts to load for Chef Infra Client.
 
 ``client_key``
-   The location of the file that contains the client key. Default value: ``/etc/chef/client.pem``. For example:
+   Warning: This config item should always be specified as an absolute path. The location of the file that contains the client key. Default value: ``/etc/chef/client.pem``. For example:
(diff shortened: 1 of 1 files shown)


IanMadd:
I don't think we need the warning sentence. It would be easier to just tell the user to specify this location as an absolute path. For example: 

> The location of the file that contains the client key, specified as an absolute path.

Does that sound good @sean-horn?

sean-horn:
I think no warning sentence the way you describe works, yes, @IanMadd

IanMadd (approved):
tenor-230603024

PR #2077, opened 2019-08-16
Add detail about chef-client package contents

sean-horn:

### Description

dtags update/replicate of chef_client_summary tag

Added more detail about what is in a chef-client package and where those contents live on the system with the package installed.

### Check List

- [X] Spell Check
- [X] Local build
- [X] Examine the local build
- [ ] All tests pass

### Issues Resolved

Customer asking for more detail about what's in the chef-client package.

### DIFF
--- a/chef_master/source/chef_client_overview.rst
+++ b/chef_master/source/chef_client_overview.rst
@@ -15,7 +15,8 @@ Chef Infra Client Overview
 
      - .. tag chef_client_summary
 
-       Chef Infra Client is an agent that runs locally on every node that is under management by Chef Infra Server. When a Chef Infra Client is run, it will perform all of the steps that are required to bring the node into the expected state, including:
+       Chef Infra Client is an agent that runs locally on every node that is under management by Chef Infra Server. The chef-client install package includes a self-contained ruby install
+       and the chef-client code, which runs in the included ruby interpreter runtime. All components of the chef-client live in /opt/chef or c:\\opscode. When a Chef Infra Client is run, it        will perform all of the steps that are required to bring the node into the expected state, including:
 
        * Registering and authenticating the node with Chef Infra Server
        * Building the node object
--- a/chef_master/source/chef_client_overview.rst
+++ b/chef_master/source/chef_client_overview.rst
@@ -15,7 +15,7 @@ Chef Infra Client Overview
 
      - .. tag chef_client_summary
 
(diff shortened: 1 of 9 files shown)


chef_master/source/chef_client_overview.rst
```diff
@@ -15,7 +15,8 @@ Chef Infra Client Overview
 
      - .. tag chef_client_summary
 
-       Chef Infra Client is an agent that runs locally on every node that is under management by Chef Infra Server. When a Chef Infra Client is run, it will perform all of the steps that are required to bring the node into the expected state, including:
+       Chef Infra Client is an agent that runs locally on every node that is under management by Chef Infra Server. The chef-client install package includes a self-contained ruby install
```
tas50: I'm not sure about leading with this. Our first introduction to what the client is shouldn't be that it ships with Ruby and everything else. That's getting pretty deep into the weeds right off the bat.

kagarmoe: It is in the weeds and we need a specific place for "Chef Infra Client contains the following: X, Y, Z, ruby install". Who else thinks the content of this "overview" page is all over the place?

kagarmoe: @tas50 @sean-horn Can you give me a list of the important things included in the chef-client? For example, the self-contained ruby install. The versions of said things might also be important. Thanks!

sean-horn: @kagarmoe Feel free to place it wherever makes sense. This was a sales/field request. I don't think we should go into versions(they change) nor any deeper than what I've written there. Folks don't necessarily need/want to know. Those that do need to know will get there fairly easily.

However, it could look something like 
* A chef-client OS package produced by Omnibus https://github.com/chef/omnibus encapsulates the following and more
  * A Ruby interpreter built from source (not in windows)
  * Various other software dependencies from https://github.com/chef/omnibus-software included in the build

kagarmoe: I'm putting this onto the "Install ChefDK" page.

kagarmoe:
Reworded a bit, so the path and content ChefDK lives only on the install page. Hopefully, users will find it with relevant content, not mixed in with overview content.

sean-horn:
I like it, @kagarmoe Thanks! :shipit:

IanMadd (approved):
tenor-150395026

PR #2085, opened 2019-08-21
Update config_rb_server.rst

lbakerchef:
``postgresql['sslmode']``
	SSL encryption mode between Chef Server and PostgreSQL.  Valid settings are 'disable' (off) and 'require' (on).  Default value: 'disable'.

### DIFF
--- a/chef_master/source/config_rb_server.rst
+++ b/chef_master/source/config_rb_server.rst
@@ -47,7 +47,7 @@ The following settings are typically added to the server configuration file (no 
 
 .. end_tag
 
-SSL Protocols
+NGINX SSL Protocols
 -----------------------------------------------------
 .. tag server_tuning_nginx
 
@@ -288,3 +288,6 @@ The following setting is often modified from the default as part of the tuning e
       550 = 350 + [(4 - 2) * (25 * 4)]
 
 .. end_tag
+
+``postgresql['sslmode']``
+   SSL encryption mode between the Chef Infra Server and PostgreSQL.  Valid settings are ``'disable'`` and ``'require'``.  Default value: ``'disable'``. 


IanMadd:
<s>Hi @lbakerchef, can you add your [DCO sign-off ](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco) to this PR. Thanks!</s> DONE

PrajaktaPurohit (approved):
Can be merged after Chef-server 13.1.x is released.

markan (changes requested):
Are there any configuration values that affect what SSL keys chef-server trusts (I suspect we've added some functionality around that as part of our general hardening effort on internal service calls).
If those exist, we should document them.

markan (approved):
Looks reasonable, but I might consider whether the SSL postgres options need to be documented in the performance tuning section or not.

PrajaktaPurohit:
@markan Adding https://github.com/chef/chef-server/issues/1850 to capture some of your input.

lbakerchef:
Good catch on the performance tuning section.  Can't remember the rationale for why that was put there.

chef_master/source/config_rb_server.rst
```diff
@@ -288,3 +288,6 @@ The following setting is often modified from the default as part of the tuning e
       550 = 350 + [(4 - 2) * (25 * 4)]
 
 .. end_tag
+
+``postgresql['sslmode']``
+   SSL encryption mode between Chef Server and PostgreSQL.  Valid settings are ``'disable'`` (off) and ``'require'`` (on).  Default value: ``'disable'``.
```
IanMadd:
```suggestion
   SSL encryption mode between the Chef Infra Server and PostgreSQL.  Valid settings are ``'disable'`` and ``'require'``.  Default value: ``'disable'``. 
```

PR #2092, opened 2019-08-27
Reformat release notes so they can be properly converted to md

IanMadd:

### Description

This PR reformats the release notes page. The existing formatting works in Sphinx but it isn't strictly correct. This creates problems when it's converted into markdown using pandoc.

There were a few problems:

* Lists were indented when they should have been aligned to the left of the page.
* Nested list items should only be indented by two spaces.
* Definitions incorporated into a list look fine in Sphinx but get a bit crazy when they're converted to markdown. Pandoc converts these definitions into blockquote text. 
* Some paragraphs that are part of a list item are indented further than they shouldn't be. Pandoc also converts those to blockquote text.
* List tables without defined widths got a little weird when converted to markdown. 
* Some text in code blocks was indented too far.

### Check List

- [ ] Spell Check
- [X] Local build
- [X] Examine the local build
- [ ] All tests pass


mjingle (approved):
tenor-224151307

PR #2095, opened 2019-08-29
Remove documentation on migrating from Ohai 6

tas50:
Ohai 7 came out in Chef 11, which was a LONG LONG time ago. There's no need to be describing this upgrade process at this point.


IanMadd (approved):
tenor-133538184

PR #2096, opened 2019-08-29
Note that the /ohai directory requires Chef Server 12.18+

tas50:
This is a fairly new feature and users tend to not upgrade their
servers. They're going to be really confused if they're on Chef Infra
Client 15 and this doesn't work. It's dependent on a new-ish Chef Infra
Server.


IanMadd (approved):
tenor-22900074

PR #2103, opened 2019-09-08
Use the correct name for the openssl_x509_certificate example

tas50:
Technically the name here works, but that's the legacy name we don't
want folks to use anymore.


IanMadd (approved):
tenor-164187568

PR #2104, opened 2019-09-08
Remove mention of providers

tas50:
This is serious inside baseball. There's no reason to tell anyone how
our resource to provider mapping works.


IanMadd (approved):
tenor-112351739

PR #2106, opened 2019-09-10
Add more resource examples

tas50:
We didn't add these when we shipped a lot of these resources. They're from the various cookbooks we shipped in the client. I also added the boilerplate text to the generator code in chef/chef so the formatting is now standardized there.


kagarmoe (approved):
tenor-113070393

Issue #2108, opened 2019-09-10
Chef Infra Server upgrade instructions are missing link to Chef-Backend upgrade page


teknofire:
### User Story
As a SysAdmin I would like to have instructions on how to upgrade my Chef Backend HA system to the latest version.

## Description
With the EOL of the Chef DRBD HA solution, we removed the HA upgrade instructions from https://docs.chef.io/upgrade_server.html#high-availability and didn't add or keep instructions for how to upgrade users with Chef-Backend HA.

There is a separate page that includes those instructions here: https://docs.chef.io/upgrade_server_ha_v2.html that should be linked to from that section.

## Chef Version
Chef-Backend all versions

## Platform Version
<! Tell us which Operating System distribution and version you're running Chef on. >

### Definition of Done
User can find instructions on how to upgrade Chef-Backend HA installations from both https://docs.chef.io/upgrade_server.html#high-availability and https://docs.chef.io/upgrade_server_ha_v2.html

PR #2110, opened 2019-09-10
Update Chef Backend Installation/Migration

kagarmoe:
### Description

Redirects links on `install-server.html` to overview, directs drbd/keepalived users to migration video.

### Issues Resolved

https://github.com/chef/chef-web-docs/issues/2108

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


IanMadd (approved):
tenor-69254183

chef_master/source/install_server_ha.rst
```diff
@@ -122,7 +122,7 @@ These instructions assume you are using the following versions or newer:
 - Chef Server  : 12.5.0
 - Chef Backend : 0.8.0
```
teknofire, line 123: These versions are very old, should we update them or possibly word this section differently so we don't have to keep them up to date?

chef_master/source/install_server_ha.rst
```diff
@@ -273,7 +276,7 @@ For each additional frontend node you wish to add to your cluster:
 
    .. code-block:: bash
 
-     	$ chef-backend-ctl gen-server-config <FE_NAME-FQDN> > chef-server.rb.<FE_NAME>
```
teknofire, line 276: Can we change the title of this section to be `Step 6: Adding More Frontends`. This is really a required step in the setup process but because it's not stated as such many users skip following these instructions when setting up a second or third FE node.

This results in an installation where only the FE node that last ran `chef-server-ctl reconfigure` can actually talk to the Backend nodes. Most users don't notice this because they have a load-balancer in front that automatically marks the failing nodes as unavailable and sends all traffic to the working node so no error is ever displayed.

It might also be good to add a note to Step 5 that state something like
> If you have already configured a FE node skip to Step 6 for configuring additional ones.

chef_master/source/upgrade_server.rst
```diff
@@ -73,12 +73,13 @@ To upgrade to Chef Server 12 from a standalone Chef Server, do the following:
 
       $ chef-server-ctl cleanup
 
-High Availability
+High Availability: Chef Backend
 -----------------------------------------------------
-This section describes the upgrade process from a high availability configuration. The upgrade process will require downtime equal to the amount of time it takes to stop the server, run dpkg or RPM Package Manager, and then upgrade the server.
+The Chef Infra Server can operate in a high availability configuration that provides automated load balancing and failover for stateful components in the system architecture
+To upgrade from Chef Infra Server to Chef Backend, follow the steps in `High Availability: Upgrade to Chef Backend 2 </upgrade_server_ha_v2.html>`_.
```
teknofire: This wording doesn't seem quite right, this sounds more like it's talking about migrating from a Chef Infra Server Standalone installation to a Chef Backend configuration.  Which would be the same steps outlined for the DRBD/keepalived configuration via https://blog.chef.io/2018/04/06/best-practices-for-migrating-your-chef-server/

chef_master/source/upgrade_server.rst
```diff
@@ -73,12 +73,13 @@ To upgrade to Chef Server 12 from a standalone Chef Server, do the following:
...
+The upgrade process will require downtime equal to the amount of time it takes to stop the server, run dpkg or RPM Package Manager, and then upgrade the server.
 
-The DRBD/Keepalived based HA system is no longer supported in Chef Infra Server 13. If a high availablity system is needed,
-backup and restore to a Chef Backend based installation.
+To upgrade Chef Backend from an existing DRBD/keepalived configuration, see the `Best Best Practices for Migrating Your Chef Server <https://blog.chef.io/2018/04/06/best-practices-for-migrating-your-chef-server/>`__ webinar from the `Chef Blog <https://blog.chef.io/>`__
```
teknofire, line 83: I would call this a migration and not an upgrade from DRBD to Chef-Backend.  There is no in-place upgrade mechanism and the customer will need to deploy completely new infrastructure and migrate their data to the new Chef-Backend cluster installation.

chef_master/source/upgrade_server_ha_v2.rst
```diff
@@ -139,4 +148,10 @@ Check the status of the cluster:
 
 .. code-block:: bash
 
-   % chef-backend-ctl status
+     chef-backend-ctl status
+
+DRBD/Keepalived HA to Chef Backend 2.x
+=====================================================
+
+For a guide to upgrading to Chef Backend from DRBD see the `Best Best Practices for Migrating Your Chef Server <https://blog.chef.io/2018/04/06/best-practices-for-migrating-your-chef-server/>`__ webinar from the `Chef Blog <https://blog.chef.io/>`__
```
teknofire: Change `upgrading` to `migrate`, maybe a small note as well that states that DRBD configurations are EOL and no longer supported.

chef_master/source/upgrade_server.rst
teknofire: To upgrade the versions of the Chef-Backend nodes they would want to follow the instructions at `/upgrade_server_ha_v2.html`. 

There is also a separate section in that page for upgrading the FE nodes for a Chef-Backend HA cluster on the install server pages here: `https://docs.chef.io/install_server_ha.html#upgrading-chef-infra-server-on-the-frontend-machines`.  This is linked to in step 8 on https://docs.chef.io/upgrade_server_ha_v2.html#chef-backend-1-x-to-2-x-upgrade but has the incorrect URL.  I'll have some more comments on that page for this.

chef_master/source/upgrade_server_ha_v2.rst
```diff
@@ -30,18 +31,18 @@ Chef Backend 1.x to 2.x Upgrade
...
+
      .. code-block:: bash
 
         yum install PATH_TO_FILE.rpm
 
-   * Debian and Ubuntu:  
-     
+   Debian and Ubuntu:
+
      .. code-block:: bash
```
teknofire, line 46: Github won't let me comment directly on this section as part of a review, but in `Step 8` below the instructions have an incorrect link: 

`https://docs.chef.io/install_server_ha.html#upgrading-chef-server-on-the-frontend-machines` 
should be
`https://docs.chef.io/install_server_ha.html#upgrading-chef-infra-server-on-the-frontend-machines`

chef_master/source/install_server_ha.rst
```diff
@@ -273,7 +276,7 @@ For each additional frontend node you wish to add to your cluster:
...
+       chef-backend-ctl gen-server-config <FE_NAME-FQDN> > chef-server.rb.<FE_NAME>
 
 #. Copy it to ``/etc/opscode`` on the new frontend node.
```
teknofire, line 282: Also, after these instructions is another section `Upgrading Chef Infra Server on the Frontend Machines`

This feels like it should be more part of the `upgrade_server_ha_v2.rst` documentation and also requires you to jump to another page with the standalone upgrade instructions.  Makes it very easy for users to miss this one important step that needs to be done before they upgrade the FE nodes.

Then on item `#3` there is a second link to those instructions that's not formatted correctly and results in a bad url: https://docs.chef.io/standaloneupgradeprocess/upgrade_server.html#standalone

teknofire (commented):
There were a few items I found that might be worth taking a look at.  Mostly some other issues I noticed with the upgrade instructions as I was taking a detailed read through.

chef_master/source/upgrade_server_ha_v2.rst
kagarmoe: done

chef_master/source/upgrade_server.rst
kagarmoe: Agree

chef_master/source/upgrade_server_ha_v2.rst
kagarmoe: done

chef_master/source/upgrade_server_ha_v2.rst
kagarmoe: done

chef_master/source/install_server_ha.rst
kagarmoe: Fixes the link. Keeping the section, we can revisit later.

PR #2117, opened 2019-09-14
Remove more deprecated metadata

tas50:
We've greatly simplified the metadata.rb file by removing content we
never actually consumed anywhere.


IanMadd (approved):
tenor-12638694

PR #2118, opened 2019-09-15
Load current value

MarkGibbons:
### Description

Some of the examples used to document load_current_values use undocumented effects or don't work.  This pull request builds on pr #1854 to clarify some of the options and includes examples that work.

### Definition of Done

PR merged

### Issues Resolved

None 

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


chef_master/source/custom_resources.rst
```diff
@@ -577,14 +577,14 @@ The ``converge_if_changed`` method may be used multiple times. The following exa
 
 .. code-block:: ruby
 
-   property :path, String, name_property: true
+   property :path, String
    property :content, String
    property :mode, String
 
-   load_current_value do
-     if ::File.exist?(new_resource.path)
```
tas50, line 572: I think the simpler fix here is just to remove new_resource. I believe you can just access these as `path`

chef_master/source/custom_resources.rst
```diff
@@ -544,7 +544,7 @@ For example, a custom resource defines two properties (``content`` and ``path``)
 .. code-block:: ruby
 
    property :content, String
-   property :path, String, name_property: true
+   property :path, String, identity: true
```
tas50: Is there a reason for this? Identity is pretty much legacy analytics stuff at this point.

chef_master/source/custom_resources.rst
MarkGibbons: Accessing them as path fails.  You get path == nil.  We either have to set one of the three attributes for the property or use a block argument in order to see the properties.

chef_master/source/custom_resources.rst
MarkGibbons: It's one of the documented ways to access the property from inside of a load_current_value block so I thought I would use it as an illustration. I think of this as less obscure than desired_state false but I don't really understand the purpose of either desired_state or identity.

MarkGibbons: Everything can't be the name_property.

chef_master/source/custom_resources.rst
MarkGibbons: Tested with chef 14 and chef 15

MarkGibbons:
Could I get another look at this one?  My testing does not agree with the review.  I think getting this one merged would improve matters a lot.  Most people I know have not adopted converge if changed because the examples don't work and the block argument doc is really unclear.

My example shows that unless something special is done the property values are not available in load_current_value. 

**Recipe**

testbook_org3 'do it' do
    path '/tmp/a'
    content 'test1'
end

**Custom resource in the testbook cookbook**
property :path, String
property :content, String

load_current_value do
  if ::File.exist?(path)
    content IO.read(path)
  end
end

action :create do
  converge_if_changed :content do
    IO.write(new_resource.path, new_resource.content)
  end
end

**Results**
         ================================================================================
           Error executing action `create` on resource 'testbook_org3[do it]'
           ================================================================================

           TypeError
           ---------
           no implicit conversion of nil into String

           Cookbook Trace:
           ---------------
           /tmp/kitchen/cache/cookbooks/testbook/resources/org3.rb:5:in `exist?'
           /tmp/kitchen/cache/cookbooks/testbook/resources/org3.rb:5:in `block in class_from_file'

           Resource Declaration:
           ---------------------
           # In /tmp/kitchen/cache/cookbooks/testbook/recipes/default.rb

            18: testbook_org3 'do it' do
            19:   path '/tmp/a'
            20:   content 'test1'
            21: end
            22:

chef_master/source/custom_resources.rst
MarkGibbons: I changed the example to use name_property.

chef_master/source/custom_resources.rst
kagarmoe: I'm taking @MarkGibbons side on this; he's worked with it most recently.

tas50 (approved):

PR #2126, opened 2019-09-17
Replace Foodcritic with Cookstyle in more places

tas50:
For anyone starting today they should skip over Foodcritic entirely as we're in the process of removing it from our tooling.


IanMadd (approved):
tenor-212014440

PR #2127, opened 2019-09-17
Remove the dedicated examples page

tas50:
This is one of our stranger pages. It's just a pile of misc Chef Infra
code that's really not helpful and requires us to maintain some quiet
funky tagging. I think the idea was to show people what Chef Infra code
looked like, but at this point it's just pointing people towards funky /
dated code snippets they shouldn't replicate.


IanMadd (approved):
tenor-189335580

PR #2135, opened 2019-09-18
Remove some very dated custom resource notes

tas50:
This whole page needs to go away, but this is a start.


IanMadd (approved):
tenor-106449198

PR #2138, opened 2019-09-19
add snap_package resource to list

### LINKED ISSUE #2136: update documentation with snap_package resource
### Describe the Enhancement:
I would like to know how to properly use the snap_package resource but do not find any documentation [here](https://docs.chef.io/resource_package.html) on how to do so.

### Describe the Need:
An administrator might want to know how configure a system using this resource.

I personally would like to configure my system with this resource because there are certain packages which are available both under apt and snap and I would like to choose the snap alternative.

### Current Alternative
Instead of referring to the documentation, one can look at how the snap_package resource is implemented in order to figure out how it works.

jlosito:
### Description

Links to snap package resource in the package resource page. 

### Issues Resolved
fixes #2136


IanMadd (approved):
Hi @jlosito, thanks for your contribution!
tenor-226704546

PR #2144, opened 2019-09-25
More changes to environments

tas50:
Explain cookbook version pinning

Remove Attribute Persistence which should only be in the attributes doc

Remove the () from node.environment. This isn't javascript

Move the chef solo instructions to their own category

Make an initial mention of cookbook pinning. This is a super important feature that all our large customers do and we don't even call it out here. Kinda weird

Remove the odd section explaining what Ruby is. We have that earlier on and we don't have a similar section for what JSON is. Eventually we need to get hover overs or aggressive links for this kinda thing. Right now though this just bloats the doc and detracts from the real content.

### DIFF
--- a/chef_master/source/api_chef_server.rst
+++ b/chef_master/source/api_chef_server.rst
@@ -2827,7 +2827,7 @@ The response is similar to:
 -----------------------------------------------------
 .. tag environment
 
-An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. Every organization begins with a single environment called the ``_default`` environment, which cannot be modified (or deleted). Additional environments can be created to reflect each organization's patterns and workflow. For example, creating ``production``, ``staging``, ``testing``, and ``development`` environments. Generally, an environment is also associated with one (or more) cookbook versions.
+An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. Every organization begins with a single environment called the ``_default`` environment, which cannot be modified (or deleted). Additional environments can be created to reflect each organization's patterns and workflow. For example, creating ``production``, ``staging``, ``testing``, and ``development`` environments.
 
 .. end_tag
 
--- a/chef_master/source/api_chef_server.rst
+++ b/chef_master/source/api_chef_server.rst
@@ -2827,7 +2827,7 @@ The response is similar to:
 -----------------------------------------------------
 .. tag environment
 
-An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. This mapping is accomplished by setting attributes and pinning cookbooks at the environment level. With environments, cookbook configurations can be changed depending on the system's designation. For example you may want to change the URL to a database server depending on if a system is in a staging or a production environment. Environments also allow organizations to move new cookbook releases from staging to production with confidence by stepping releases through testing environments before entering production.
+An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. This mapping is accomplished by setting attributes and pinning cookbooks at the environment level. With environments, you can change cookbook configurations depending on the system's designation. For example, by designating different staging and production environments, you can then define the correct URL of a database server for each environment. Environments also allow organizations to move new cookbook releases from staging to production with confidence by stepping releases through testing environments before entering production.
 
 .. end_tag
 
--- a/chef_master/source/api_chef_server.rst
+++ b/chef_master/source/api_chef_server.rst
@@ -2827,7 +2827,7 @@ The response is similar to:
 -----------------------------------------------------
 .. tag environment
 
-An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. Every organization begins with a single environment called the ``_default`` environment, which cannot be modified (or deleted). Additional environments can be created to reflect each organization's patterns and workflow. For example, creating ``production``, ``staging``, ``testing``, and ``development`` environments.
+An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. This mapping is accomplished by setting attributes and pinning cookbooks at the environment level. With environments, cookbook configurations can be changed depending on the system's designation. For example you may want to change the URL to a database server depending on if a system is in a staging or a production environment. Environments also allow organizations to move new cookbook releases from staging to production with confidence by stepping releases through testing environments before entering production.
 
 .. end_tag
 
--- a/chef_master/source/environments.rst
+++ b/chef_master/source/environments.rst
@@ -5,21 +5,21 @@ About Environments
 
 .. tag environment
 
-An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. Every organization begins with a single environment called the ``_default`` environment, which cannot be modified (or deleted). Additional environments can be created to reflect each organization's patterns and workflow. For example, creating ``production``, ``staging``, ``testing``, and ``development`` environments. Generally, an environment is also associated with one (or more) cookbook versions.
+An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. Every organization begins with a single environment called the ``_default`` environment, which cannot be modified (or deleted). Additional environments can be created to reflect each organization's patterns and workflow. For example, creating ``production``, ``staging``, ``testing``, and ``development`` environments.
 
 .. end_tag
 
 The _default Environment
 =====================================================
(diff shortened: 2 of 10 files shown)


lamont-granquist (approved):

chef_master/source/api_chef_server.rst
```diff
@@ -2827,7 +2827,7 @@ The response is similar to:
 -----------------------------------------------------
 .. tag environment
 
-An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. Every organization begins with a single environment called the ``_default`` environment, which cannot be modified (or deleted). Additional environments can be created to reflect each organization's patterns and workflow. For example, creating ``production``, ``staging``, ``testing``, and ``development`` environments. Generally, an environment is also associated with one (or more) cookbook versions.
+An environment is a way to map an organization's real-life workflow to what can be configured and managed when using Chef Infra. Every organization begins with a single environment called the ``_default`` environment, which cannot be modified (or deleted). Additional environments can be created to reflect each organization's patterns and workflow. For example, creating ``production``, ``staging``, ``testing``, and ``development`` environments.
```
kagarmoe: An environment is a way to map an organization's real-life workflow TO WHAT can be configured and managed when using Chef Infra.

Can we find a more useful phrasing? TO WHAT is obscure. 

Environments are useful for mapping an organization's real-life workflows to Chef Infra configurations and management definitions.

This is awkward, but a beginning.

chef_master/source/environments.rst
```diff
@@ -5,21 +5,21 @@ About Environments
...
 
 .. end_tag
 
 The _default Environment
 =====================================================
-Every organization must have at least one environment. Every organization starts out with a single environment that is named ``_default``, which ensures that at least one environment is always available to the Chef Infra Server. The ``_default`` environment cannot be modified in any way. Nodes, roles, run-lists, cookbooks (and cookbook versions), and attributes specific to an organization can only be associated with a custom environment.
+Every Chef Infra Server organization must have at least one environment. Each organization starts out with a single environment that is named ``_default``. The ``_default`` environment cannot be modified in any way. Nodes, roles, run-lists, cookbooks (and cookbook versions), and attributes specific to an organization can only be associated with a custom environment.
```
kagarmoe:
```suggestion
Every Chef Infra Server organization must have at least one environment. Each organization starts out with a single ``_default`` environment. The ``_default`` environment cannot be modified in any way. Nodes, roles, run-lists, cookbooks (and cookbook versions), and attributes specific to an organization can only be associated with a custom environment.
```

chef_master/source/environments.rst
```diff
@@ -5,21 +5,21 @@ About Environments
...
 
 Environment Attributes
```
kagarmoe, line 17: This section comes out of the blue. ELIF why we're suddenly discussing attributes.

chef_master/source/environments.rst
```diff
@@ -97,33 +92,21 @@ Attribute precedence, when viewed as a table:
 
 .. end_tag
 
+Cookbook Pinning
+=====================================================
+
+Cookbook versions can be pinned per evironment. This allows for the controlled rollout of new cookbook releases through environments for testing before new versions reach production environments. See the environment format examples below for the cookbook pinning syntax.
```
kagarmoe:
```suggestion
Cookbook versions can be pinned in each environment, which allows you to control the rollout of new cookbook releases through successive testing environments before releasing new cookbook versions into production environments. See the environment format examples below for the cookbook pinning syntax.
```

chef_master/source/environments.rst
```diff
@@ -97,33 +92,21 @@ Attribute precedence, when viewed as a table:
...
+
 Environment Formats
 =====================================================
-Environment data may be stored in two formats:
+
+Environments may be stored on disk (any in source control) in two formats:
```
kagarmoe, line 104:
```suggestion
Store environments on disk (in any source control) in one of two formats:
```

chef_master/source/environments.rst
```diff
@@ -382,8 +364,37 @@ A node is considered to be associated with an environment when the ``chef_enviro
   .. note:: After the environment has been set via bootstrap, the environment is set in the client.rb file and may not be modified using the Chef management console or the ``edit`` argument of the ``knife node`` subcommand.
 * By setting the ``environment`` configuration entry in the client.rb file ; when Chef Infra Client runs, it will pick up the value and then set the ``chef_environment`` attribute of the node
 
-Set using chef-solo
+Move Nodes
 -----------------------------------------------------
+Nodes can be moved between environments, such as from a "dev" to a "production" environment by using the ``knife exec`` subcommand. For example:
```
kagarmoe:
```suggestion
Use the ``knife exec`` subcommand to move nodes between environments, such as from a "dev" to a "production" environment. For example:
```

chef_master/source/environments.rst
```diff
@@ -5,21 +5,21 @@ About Environments
...
 
 Environment Attributes
 =====================================================
 .. tag environment_attribute
 
-An attribute can be defined in an environment and then used to override the default settings on a node. When an environment is applied during a Chef Infra Client run, these attributes are compared to the attributes that are already present on the node. When the environment attributes take precedence over the default attributes, Chef Infra Client applies those new settings and values during a Chef Infra Client run.
+Attributes can be defined in an environment and then used to override the default attributes in a cookbook. When an environment is applied during a Chef Infra Client run, environment attributes are compared to the attributes that are already present on the node. When the environment attributes take precedence over the default attributes, Chef Infra Client applies those new settings and values during a Chef Infra Client run.
```
kagarmoe, line 21: Defining attributes in an environment lets you override the default attributes in a cookbook. When a Chef Infra Client run applies an environment, the client compares the environment attributes to the attributes that are already defined on the node [IN THE COOKBOOK? WHY ARE WE SUDDENLY TALKING ABOUT NODES?] Environment attributes set to ``override`` take precedence over the default attributes set on the node by cookbooks. Chef Infra Client applies the ``override`` environment attributes and values during a Chef Infra Client run.

...or something like this

chef_master/source/environments.rst
```diff
@@ -5,21 +5,21 @@ About Environments
...
 
-An environment attribute can only be set to be a default attribute or an override attribute. An environment attribute cannot be set to be a ``normal`` attribute. Use the ``default_attribute`` and ``override_attribute`` methods in the Ruby DSL file or the ``default_attributes`` and ``override_attributes`` hashes in a JSON data file.
+Environment attributes can be set to either ``default`` attribute level or an ``override`` attribute level.
```
kagarmoe, line 23: Maybe an example would help?

chef_master/source/environments.rst
```diff
@@ -53,16 +53,11 @@ There are two types of attributes that can be used with environments:
 
        .. end_tag
 
-Attribute Persistence
+Attribute Precedence
```
kagarmoe, line 57: "Precedence" is a much better word! 👍

chef_master/source/environments.rst
```diff
@@ -53,16 +53,11 @@ There are two types of attributes that can be used with environments:
...
 -----------------------------------------------------
-.. tag node_attribute_persistence
-
-All attributes except for normal attributes are reset at the beginning of a Chef Infra Client run. Chef Infra Client rebuilds these attributes using automatic attributes collected by Ohai at the beginning of each Chef Infra Client run, and then uses default and override attributes that are specified in cookbooks, roles, environments, and Policyfiles. All attributes are then merged and applied to the node according to attribute precedence. The attributes that were applied to the node are saved to the Chef Infra Server as part of the node object at the conclusion of each Chef Infra Client run.
 
-.. end_tag
+Environments are one of several locations where attributes can be applied in Chef Infra. It's important to understand how the precedence level works in order to understand, which attributes will be applied when Chef Infra Client runs.
```
kagarmoe, line 60: Environments are one of several different places [FILES?] for defining attributes in Chef Infra. Attributes are defined in cookbook attribute files, recipes, environments, roles, and also in Ohai.  Attribute _precedence_ explains why an attribute defined in one file wins out over another during a Chef Infra Client run and gets applied to a node.

kagarmoe: I think we should reverse the order of the attribute precedence list, so #1 is the Ohai automatic attribute.

kagarmoe (commented):
Mostly exploratory comments and revisions.

chef_master/source/environments.rst
tas50: We said may because it's entirely optional to store them on disk. It's the right thing to do, but still optional.

kagarmoe: I'll think about different words.

kagarmoe (approved):

PR #2154, opened 2019-10-03
Rework some DK docs

tas50:
Gets these ready to be workstation docs down the line.


kagarmoe (approved):
tenor-36843707

PR #2156, opened 2019-10-03
Update dk_windows.rst

douglax:
The original doc read to "navigate to Github downloads", but the link would take you to the Git project download page, not the Github download (which doesn't exist, or -at least- is not accurate). I just replaced Github with Git

### Description

It provides a more accurate description of the site where git can be downloaded

### Issues Resolved

None formal issues, just a difference in labeling

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/chef_master/source/dk_windows.rst
+++ b/chef_master/source/dk_windows.rst
@@ -138,7 +138,7 @@ Chef support in editors:
 
 Install and Optimize Git for Chef Development
 -------------------------------------------------------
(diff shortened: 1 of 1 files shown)


IanMadd:
Hi @douglax, thanks for your contribution. Could you add "Obvious fix" to your commit comment. See our Obvious Fix policy here: https://github.com/chef/chef/blob/master/CONTRIBUTING.md#chef-obvious-fix-policy

tas50:
@douglax This part of the doc isn't actually accurate as git ships with DK now. I'll get that all cleaned up.

PR #2157, opened 2019-10-03
Remove mention of installing git into DK & update URLs

tas50:
Also update the example output to be a year newer.


IanMadd (approved):
tenor-105230814

PR #2158, opened 2019-10-03
Better describe that we'd like folks to stop using definitions

tas50:
They're not deprecated, but folks really need to stop using them. Also the example at the bottom wasn't valid chef code.


IanMadd (approved):
tenor-30183722

PR #2160, opened 2019-10-03
Remove Chef books that are 5+ years old

tas50:
We want to make sure we're directing folks to books that are going to describe current best practices. A lot of these books were great when written, but are very dated at this point. This removes everything that's about 5+ years old and collapses the list into a single set of books w/o recommendations. We haven't reviewed everything that's out there so we can't say what's best.


IanMadd (approved):
tenor-20606508

PR #2161, opened 2019-10-03
Update external links to avoid redirects

tas50:
Let's keep the links current here.


IanMadd (approved):
tenor-4514148

PR #2168, opened 2019-10-10
Fix incorrectly formatted inline code

IanMadd:

This adds a second set of back ticks to some strings of inline code in the release_notes_server.rst page. Some of the inline code strings weren't formatting correctly without them.

### Check List

- [ ] Spell Check
- [X] Local build
- [X] Examine the local build
- [ ] All tests pass


kagarmoe (approved):
tenor-88316484

PR #2200, opened 2019-11-26
Update example for multi string registry_key 

stuartpreston:
### Description

The :multi_string registry key type accepts a string array and is idempotent when used this way (unlike the previous example which uses nul terminated strings and is not idempotent). This small update corrects the example.

### Definition of Done

Updated example, including corrected style

### Issues Resolved

N/A - updating after observing internal discussion in Slack.

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


IanMadd (approved):
tenor-93358010

PR #2214, opened 2019-12-31
Fix header level in DSL doc

tas50:
This was under Windows before.


IanMadd (approved):
tenor-84629603

PR #2216, opened 2020-01-03
Move the products that deprecated at EOY 2019 as planned

btm:
Signed-off-by: Bryan McLellan <btm@loftninjas.org>

### DIFF
--- a/chef_master/source/versions.rst
+++ b/chef_master/source/versions.rst
@@ -80,17 +80,9 @@ Use of these and later versions of these distributions must be in accordance wit
      - 4.x
      - GA
      - n/a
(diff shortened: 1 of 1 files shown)


susanev:
i noticed that Chef Provisioning is in the end-of-life section but the lifecycle status says _Deprecated_, is that accurate?

kagarmoe:
Docs will do this once we have the PR for removing the content finished.

mjingle (approved):

btm:
I rebased this and removed a couple of EOL dates that we set preemptively that are fast approaching that will be changed.

PR #2234, opened 2020-01-23
Add new Chef Infra Client 15.7 resource properties

tas50:
Also sync over formatting from the auto generated docs so more things match and diffs are easier.


IanMadd (approved):
tenor-45385928

PR #2237, opened 2020-01-24
Add Chef Infra Client 15.5 - 15.7 release notes

tas50:
We're behind here. Let's catch up a bit.


IanMadd (approved):
tenor-132747734

PR #2238, opened 2020-01-25
Add DK 4.6 release notes and rework 3.3 - 4.5

tas50:
Rework these release notes to not be a giant list. It's really hard to read and even harder to get right without build failures. This matches how we do the Chef Infra Client notes. I didn't go all the way back because my brain started to hurt shuffling things around into consistent categories.


IanMadd (approved):
tenor-77461276

PR #2245, opened 2020-01-25
A bit of cleanup to the windows page

tas50:
Fix a version typo that messed up the formatting

Remove the odd section about setting the path. Setting it via the ENV['PATH'] method doesn't work the way folks would assume and the PHP cookbook that this example came from doesn't do that anymore

Remove the Azure cookbook since it is no longer maintained


kagarmoe (approved):
tenor-53651636

PR #2246, opened 2020-01-26
Remove Automate Workflow from the overview

tas50:
We shouldn't include this in the overview of Automate as it's not sold
in the standard package at this point.


kagarmoe (approved):
tenor-198206028

PR #2249, opened 2020-01-28
RFC2606 - example.com

h4ckninja:
### Description

I also think it's less confusing when using the example domains in documentation rather than specific domain names. This way people can see to fill in their own values rather than a specific hostname that may or may not exist, or that may or may not have permission to use.

### Definition of Done

Minor change.

### Issues Resolved

N/A

### Check List

- [x] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/chef_master/source/config_rb.rst
+++ b/chef_master/source/config_rb.rst
@@ -201,7 +201,7 @@ If the Chef Infra Server is configured to use HTTP, add the following settings:
 
    .. code-block:: ruby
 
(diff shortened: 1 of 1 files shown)


mjingle (approved):
Note: Because this is an obvious fix, this PR is submitted with no DCO required. 

Thank you for your contribution, @h4ckninja. We appreciate you taking time to correct this!

PR #2260, opened 2020-02-04
Remove opscode from the example config.rb

tas50:
Let's update this example for 2016.


IanMadd (approved):
tenor-223989636

PR #2266, opened 2020-02-05
Remove mention of changing install location on Mac

tas50:
You can't do that on a Mac


IanMadd (approved):
tenor-180882231

PR #2270, opened 2020-02-06
Updated deps

kagarmoe:

In the spirit of curiosity, I updated all of the deps in `requirements.txt`. It dramatically improves the build time. We probably don't want to merge this because of unintended consequences with the docs conversion.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,23 +3,23 @@
 #
 #    pip install -r requirements.txt
 #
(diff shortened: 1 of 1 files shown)


kagarmoe:
How long did the build take?
Screen Shot 2020-02-05 at 10 12 36 PM

IanMadd (approved):
It's still slower than Hugo, so:
tenor-78682779
But faster than before, so:
tenor-93017654

PR #2290, opened 2020-02-20
Remove berks test command that is no longer in berks

tas50:
This just results in an error


IanMadd (approved):
tenor-193894474

PR #2291, opened 2020-02-20
Simplify the Delivery CLI installation section

tas50:
It's part of Workstation. There isn't more to mention there. We'll automatically cleanup bin links for any old version and that would be from many many YEARS ago. Also remove the note about ancient Automate.


IanMadd (approved):
tenor-223656157

PR #2294, opened 2020-02-20
Update the wording around cookbook authoring best practices

tas50:
"platform resources" is not a thing. I know what we're trying to say
here, but that's just not a thing.


IanMadd (approved):
tenor-20456390

PR #2299, opened 2020-02-20
Replace elseif with elsif in Ruby examples

### LINKED ISSUE #2284: Replace elseif with elsif in Ruby examples
https://docs.chef.io/dsl_recipe.html#platform has `elseif` but in Ruby it's actually `elsif`. There may be other occurrences in the docs.

https://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Control_Structures#if-elsif-else_expression

tas50:
Fixes https://github.com/chef/chef-web-docs/issues/2284


IanMadd (approved):
tenor-14975613

PR #2300, opened 2020-02-20
Fix a few minor issues in the custom resource docs

tas50:
The example code didn't actually work.


IanMadd (approved):
tenor-97076455

PR #2303, opened 2020-02-20
Add a deprecation warning to the Foodcritic page

tas50:
There are a lot of users so we want to keep the docs, but we should warn folks against using it.


IanMadd (approved):
tenor-58123587

PR #2306, opened 2020-02-20
Be more clear that optional plugins need to be symbols

tas50:
We fixed this for the next Ohai, but everyone struggled with it.


stephenlauck:
👍

IanMadd (approved):
tenor-121363053

PR #2320, opened 2020-02-21
Remove all the leading $ from shellout examples

tas50:
These don't make sense when we have code blocks.


IanMadd:
Nice branch name. I approve 100%.

IanMadd (approved):
tenor-172640454

PR #2336, opened 2020-02-26
Fix appearance of shortcode

IanMadd:

### Description

Due to the idiosyncrasies of Hugo and Goldmark and how they process shortcodes, this text looks awful after its rendered. This will fix the appearance of the text on all three pages, but also creates four shortcodes on three pages where there used to be just one shortcode. I'll try to come up with a more permanent solution at some point.

### Issues Resolved

### Check List

- [ ] Spell Check
- [X] Local build
- [X] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/chef_shell.md
+++ b/content/chef_shell.md
@@ -69,7 +69,13 @@ This command has the following options:
 
     {{< warning >}}
 
-    {{% node_ctl_attribute %}}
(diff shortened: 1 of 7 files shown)


mjingle (approved):
tenor-119180012

PR #2343, opened 2020-02-26
Don't include install instructions for built-in knife commands

tas50:
These are part of Chef Workstation.


kagarmoe (approved):
tenor-187381256

PR #2347, opened 2020-02-27
Fixes for the HA upgrade doc

tas50:
Remove the odd image at the top of the page
Use a relative link within the doc so you can browse a local build without hitting prod


IanMadd (approved):
tenor-54324696

PR #2349, opened 2020-02-27
Use relative links everywhere

tas50:
This way you stay local when you're testing locally


IanMadd:
I want to yank the `.html` from the end of those relative links. It should be `/` instead. I can take care of it in a little bit.

IanMadd (approved):
tenor-193202006

PR #2380, opened 2020-03-10
Improve pushy network port documentation

atrull:
### Description

Adding to the documentation to cover some edge-case issues. Port 10003 wasn't mentioned before.

### Definition of Done

The documentation is clear and understandable and relates to reality.

### Issues Resolved

https://github.com/chef/opscode-pushy-server/issues/193

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


atrull:
If there's somewhere else in the documentation we should also document ports, I haven't found it.

atrull:
Found one.. I've extended another document file with mention of the api_port too.

PR #2383, opened 2020-03-11
Ohai plugin install

scottvidmar:
Signed-off-by: Scott Vidmar <svidmar@chef.io>

### Description

This change provides an example cookbook to install an OHai plugin on a node, and make it available without a chef-client run (ad-hoc OHai run).

### Issues Resolved

### Check List

- [X] Spell Check
- [X] Local build
- [X] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/ohai_custom.md
+++ b/content/ohai_custom.md
@@ -692,7 +692,7 @@ Installing
 Via a Cookbook
 --------------
 
-The following cookbook code will install a custom OHai plugin and make it accessible to any subsequent OHai execution or Chef run. To allow this plugin to be used by the rest of the run, this cookbook should be near the top of the runlist. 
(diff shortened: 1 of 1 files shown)


tas50 (changes requested):
I don't think this is the route that we want to suggest folks install ohai plugins with. For legacy users we have an ohai_plugin resource in in the ohai cookbook which handles the creation of the appropriate directories on both *nix and Windows plus the necessary ohai reload. For users of newer Chef Infra Server (yep server) releases they can just stick ohai plugins into an 'ohai' directory in their cookbook and the client does it all automatically.

scottvidmar:
I'm happy to make changes, but what do we recommend for Effortless? Would that apply to Kitchen, as well?

tas50:
Same should apply to effortless and Test Kitchen

PR #2397, opened 2020-03-23
Cleanup some bad conditional examples

tas50:
Use {} not do/end for ruby blocks
Make sure to ::File otherwise you're going to get a Cookstyle warning

### DIFF
--- a/layouts/shortcodes/resource_template_add_file_only_if_ruby.md
+++ b/layouts/shortcodes/resource_template_add_file_only_if_ruby.md
@@ -5,6 +5,6 @@ a file based on a template, and then use Ruby to specify a condition:
 template '/tmp/somefile' do
   mode '0755'
   source 'somefile.erb'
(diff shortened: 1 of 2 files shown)


layouts/shortcodes/resource_template_add_file_only_if_ruby.md
```diff
@@ -5,6 +5,6 @@ a file based on a template, and then use Ruby to specify a condition:
 template '/tmp/somefile' do
   mode '0755'
   source 'somefile.erb'
-  only_if do ! File.exist?('/etc/passwd') end
+  only_if { ! ::File.exist?('/etc/passwd') }
```
tas50, line 8: Also this is going to alert in the near future. Why not just use not_if here and avoid the need to negate the value returned?

phiggins (approved):

IanMadd (approved):

PR #2402, opened 2020-03-27
Fix shortcodes

IanMadd:

### Description

Fixes two things:

#### **Less than character**
For some reason the less than symbol "<" is converted to its html equivalent if it is in code blocks that are also in a shortcode. This is an issue in inline and block code. So `if x <= 100:` is turned into `if x &lt;= 100:`.

This isn't a problem if readFile and markdownify are used to read the shortcode in as opposed to the normal method of just adding the shortcode text.

So this adds a new shortcode called readFile_shortcode and then uses it to read the markdown content and convert it into html.

#### **Shortcodes in list items**
Shortcodes don't render properly if they're in a list. The text content takes up the full width of the page instead of indenting. The **install_server** and **install_chef_airgap** pages have several shortcodes in list items. The **shortcode_indent** shortcode handles them properly.

### Issues Resolved

### Check List

- [ ] Spell Check
- [X] Local build
- [X] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/chef_search.md
+++ b/content/chef_search.md
@@ -408,4 +408,4 @@ Data Bags
 
 {{% data_bag %}}
 
-{{% search_data_bag %}}
+{{< readFile_shortcode file="search_data_bag.md" >}}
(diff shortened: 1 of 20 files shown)


kagarmoe (approved):
tenor-193148472

PR #2407, opened 2020-04-04
/client end point updates

MarkGibbons:
Change the order of the endpoints to be a more hierarchical order
Update comments, condition codes and request bodies

Signed-off-by: markgibbons <mark.gibbons@nordstrom.com>

### Description

Updates the /client endpoints to match current chef server codes and request bodies.  This was a big change and I don't know how to display what it will look like after the update.  The changed needs a careful review.

I moved /client/Name/key endpoints after the /client endpoints to be consistent with placement.

Thanks for looking at it.

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


IanMadd (approved):
Thanks @MarkGibbons !

PR #2423, opened 2020-04-18
Add netlify config files

IanMadd:

### Description

Adds netlify config files.

This adds a link check plugin that will check for dead links
This removes all aliases from the Hugo build using an additional config file.

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


config_netlify.toml
```diff
@@ -0,0 +1,4 @@
+# Config settings for Netlify
+# https://discourse.gohugo.io/t/about-the-recent-redirect-loops-on-gohugo-io/13438
+
+disableAliases = true
```
schisamo, line 4: Since we are going to move over to using Netlify for deploying the production site should we just make this part of our core Hugo config in `config.toml`?

schisamo: I chatted with @IanMadd about this....until we recreate all of the aliases using an `_redirects` file we'll  want this config to only apply to the Netlify builds.

schisamo (approved):

PR #2425, opened 2020-04-20
[CHORE] Tidy the apt_package resource

damacus:
### Description

- Uses multi line yaml
- Removes new lines where they are not required

Adds markdown lint config
Adds yamllint config

### Issues Resolved

None

### Check List

- [x] Spell Check
- [x] Local build
- [x] Examine the local build
- [x] All tests pass


.yamllint
```diff
@@ -0,0 +1,7 @@
+---
+extends: default
+ignore: .travis.yml
```
tas50: We probably don't need this bit

damacus: If we're using yamlint, it always seems to find those files.

tas50: there shouldn't be a travis config in this repo though

IanMadd (approved):

PR #2440, opened 2020-04-23
WIP: Introduce the concept of a "Chef Language"

tas50:
I've always struggled with the way our docs were structured and I realized that a lot of it comes down to how we describe the cookbooks language in Chef. We use a lot of different names for it because we don't have an actual name. This PR gives it one true name: The Chef Infra Language which includes all our helpers and the out of the box resources. If you're writing a cookbook you're not writing Ruby or recipe code or Chef code, or Chef DSL, or Recipe DSL. You're writing Chef Infra Language code. Once we have this concept we can better split up how we describe part of our language from the structure of the cookbook. It's all intermixed right now in a way that's SUPER confusing. We do need to describe the parts of a cookbook, but that's not actually the language. That's more folder structure and high level concepts of how Chef Infra works.

I think this is going to give us room to expand to include the 50+ helpers that aren't documented yet, but are included in Chef Infra Client 15.4 and later. I didn't want to just shove those into the existing structure since it was just wrong.

### DIFF
--- a/archetypes/resource_index.md
+++ b/archetypes/resource_index.md
@@ -20,7 +20,7 @@ menu:
 
 resource_description:
 resource_note:
-resource_new_in:      
+resource_new_in:
 
 
 ######## Handler Types ########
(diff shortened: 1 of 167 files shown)


kagarmoe:
INTERESTING!
Have you thought about how to describe the parts-of-speech in the Chef Infra language?
How do we teach it?

kagarmoe:
I'm closing this out. We can reopen it if needed.

kmf:
Watching this

PR #2441, opened 2020-04-24
Add new resources in Chef Infra Client 16

tas50:
This also fixes a few misc bugs in the old docs.

### DIFF
--- a/content/resources/cab_package/_index.md
+++ b/content/resources/cab_package/_index.md
@@ -8,7 +8,7 @@ menu:
     title: cab_package
     identifier: chef_infra/cookbook_reference/resources/cab_package cab_package
     parent: chef_infra/cookbook_reference/resources
-    weight: 180
+
 resource_reference: true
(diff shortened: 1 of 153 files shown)


IanMadd (approved):

content/resources/cab_package/_index.md
```diff
@@ -1,35 +1,38 @@
 ---
 title: cab_package resource
 resource: cab_package
-draft: false
```
phiggins, line 4: I'm confused about this, was this intentional?

tas50: That was probably not intentional. @IanMadd does that have a default value?

IanMadd: It should be good. The default value is false.

tas50: Cool we can stop generating that then. Is there a complete list of the defaults?

PR #2443, opened 2020-04-27
Bump down headings in chef-web-docs

IanMadd:
### Description

This bumps all the headings in the docs down one level. H1 -> H2, H2 -> H3, etc...
H4 --> bold/strong (instead of H5)

Each page should have only one H1 heading which should be the title. Every other heading should be H2 or below. Currently we have many many H1 headings in the page text.

This also fixes the headings in the runbook page and the toc in the runbook page.
The problem is that the runbook page is a concatenation of several runbook subpages combined into one page. Each subpage has a title, which should be an H1, followed by subsections which are H2 down to H4. But in the runbook page, each page title has to be H2 because the runbook page title is H1, so that means every heading in the runbook subpages have to be bumped down one, but only in the runbook page. 

I also fixed a problem with the runbook toc which omits the subpage titles.

- [X] chef/chef-workstation#1126 has to be merged first
- [X] then I can update the reference to the chef-workstation hugo module, 
- [ ] then this can be merged.

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


kagarmoe (approved):
tenor-20492289

PR #2448, opened 2020-04-28
Revise the percentage smaller

tas50:
We slipped a bit from 33% to about 30%.


mjingle (approved):
tenor-164429874

PR #2449, opened 2020-04-28
Container endpoint

MarkGibbons:
### Description

Document the container endpoint and methods.  

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/api_chef_server.md
+++ b/content/api_chef_server.md
@@ -2523,8 +2523,10 @@ The `/containers` endpoint has the following methods: `GET`, `POST`.
 
 The `GET` method is used to get a list of containers.
 
-This method has no parameters.
+Note: The `/containers` endpoint is not useful outside of the Chef Infra Server code.
 
+This method has no parameters.
+e
 **Request**
 
 ``` none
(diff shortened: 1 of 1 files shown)


MarkGibbons:
I'm torn about this one.  I can't image using these endpoints and Chef may not want them documented.    /containers doesn't show up in the side index of the API documentation. I think that's because we need a blank line inserted.  If we leave the doc as is that part still needs to happen.  Cheers.

content/api_chef_server.md
```diff
@@ -2604,6 +2606,184 @@ The response is similar to:
 </tbody>
 </table>
 
+### POST
+
+The `POST` method is used to create a container.
+
```
markan, line 2710: Add here something like `NOTE: while this method is implemented, it is not supported, and use will likely break your system, or not have the effect you expect.`

content/api_chef_server.md
```diff
@@ -2604,6 +2606,184 @@ The response is similar to:
...
+/containers/NAME
+----------------
+
+The `/containers/Name` endpoint has the following methods: `DELETE`, `GET`.
+
+### DELETE
+
+The `DELETE` method is used to remove a container.
+
```
markan, line 2778: Add here something like `NOTE: while this method is implemented, it is not supported, and use will likely break your system, or not have the effect you expect.`

markan (changes requested):
Thank you for this PR. You rightly identify the API as somewhat problematic. 

There's certainly nothing secret about these APIs. However they are functionally useless, and possibly dangerous (DELETEing a container probably will break your system). They were implemented for completeness, and the routes for the ACLs naturally were built off of them.

I still think it's worthwhile documenting these. I would suggest adding a note for the POST/DELETE methods. Something like "NOTE: while this method is implemented, it is not supported, and use will likely break your system" for each of those methods.

content/api_chef_server.md
MarkGibbons: In the queue. Thanks

content/api_chef_server.md
```diff
@@ -2657,6 +2661,10 @@ The `/containers/Name` endpoint has the following methods: `DELETE`, `GET`.
 
 The `DELETE` method is used to remove a container.
 
+The `/containers/Name` endpoint has the following methods: `DELETE`, `GET`.
+
+Note: Using the `DELETE` method of the `/containers` endpoint may have unexpected effects and is likely to break your systemr. Use of this method is not supported.
```
markan:
```suggestion
Note: Using the `DELETE` method of the `/containers` endpoint may have unexpected effects and is likely to break your system. Use of this method is not supported. 
```

markan (approved):

markan:
Thanks again for this!

kagarmoe (approved):
tenor-42074873

PR #2452, opened 2020-04-28
Update the platforms we support

tas50:
We don't support i386 Debian
CentOS support mirrors RHEL support
Oracle should get aarch64
We support Ubuntu 20.04 aarch64
We support SLES 15 aarch64
We don't support S390x anymore


btm:
https://deploy-preview-2452--chef-web-docs.netlify.app/platforms/

content/platforms.md
```diff
@@ -70,17 +70,17 @@ versions for Chef Infra Client:
 </tr>
 <tr class="even">
 <td>Amazon Linux</td>
-<td><code>x86_64</code></td>
+<td><code>x86_64</code>, <code>aarch64</code> (2.x only)</td>
 <td>2013+ and 2.0</td>
 </tr>
 <tr class="odd">
 <td>CentOS</td>
-<td><code>x86_64</code></td>
```
btm, line 76: can we try adjusting the column widths for this table? the arch column is all bunched up with a ton of space to the left of it.

content/platforms.md
```diff
@@ -110,12 +110,12 @@ versions for Chef Infra Client:
 </tr>
 <tr class="even">
 <td>SUSE Enterprise Linux Server</td>
-<td><code>x86_64</code>, <code>s390x</code> (12.x only)</td>
```
btm, line 111: We should probably have an Architecture EOL table at the bottom of this page. It would also include when we dropped PowerPC support for Chef Infra Server.

content/platforms.md
tas50: I've adjusted the columns

content/platforms.md
tas50: That seems doing, but outside the scope of this. We'd need to go back and figure out when we dropped the various archs

btm: Can be separate, sure, but we're removing s390x here with no mention of where it went.

kagarmoe: Made an issue for it: #2455

kagarmoe (approved):

PR #2453, opened 2020-04-28
Make sure Red Hat is always two words

tas50:
See https://en.wikipedia.org/wiki/Red_Hat


phiggins:
There's one in here too: https://github.com/chef/chef-web-docs/blob/master/content/platforms.md#platform-end-of-life-policy

phiggins (approved):

tas50:
The platform one is handled in the platform PR. That's where I noticed this was an issue.

PR #2454, opened 2020-04-28
Trim down Ohai data with links and add new plugins

tas50:
This page was massive and it made a lot more sense to just link to the
better pages instead of duplicating part of those pages in one giant
monolith.


phiggins (approved):
Makes sense 👍

PR #2460, opened 2020-04-29
Update the platforms / architectures on the omnitruck page

tas50:
Fix the column widths to make it easier to read
Update the platforms and chef infra versions in the examples
Add i386 debian
Add Ubuntu 20.04
Add aarch64 SLES
Add aarch64 Amazon


kagarmoe (approved):
tenor-57984680

PR #2467, opened 2020-04-30
Update resource examples

IanMadd:
### Description

Update the resources Hugo templates:
- Removes trailing white spaces 
- replaces the `examples_list` property with `examples`

Replace the `examples_list` with `examples`:
- In each yaml file the `examples` property is now a concatenation of the previous `examples_list`

### Check List

- [ ] Spell Check
- [X] Local build
- [X] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/resources/apt_package/_index.md
+++ b/content/resources/apt_package/_index.md
@@ -140,17 +140,10 @@ properties_list:
   required: false
   description_list:
   - markdown: The version of a package to be installed or upgraded.
-examples_list:
-- example_heading: Install a package using package manager
-  text_blocks:
-  - code_block: |
-            apt_package 'name of package' do
-              action :install
-            end
-- example_heading: Install without using recommend packages as a dependency
-  text_blocks:
-  - code_block: |
-            package 'apache2' do
-              options '--no-install-recommends'
-            end
+examples: "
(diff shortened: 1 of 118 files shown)


content/resources/apt_package/_index.md
```diff
@@ -140,17 +140,10 @@ properties_list:
   required: false
   description_list:
   - markdown: The version of a package to be installed or upgraded.
-examples_list:
-- example_heading: Install a package using package manager
-  text_blocks:
-  - code_block: |
```
kagarmoe, line 146: Sorry that I'm coming late to this discussion. Did the literal block scalar just not work? I'm not wild about the hand curation needed in the solution.

IanMadd: The issue is that all the examples need to be in one long MD string. Currently there are resource examples in `chef/chef`, but they're one long string and there's no easy way to get them from `chef/chef` to the current Hugo template. When Tim updates the resources using his rake task, then the MD will look readable not like the jumbled mess that it is here.

tas50: Automate the world!

tas50:
Looks good to me. I can start automating all these now. Thanks @IanMadd

tas50 (approved):

PR #2477, opened 2020-05-04
Fix our code blocks to not contain a space

tas50:
These were blowing up rubocop markdown.


layouts/shortcodes/template_variables.md
```diff
@@ -43,13 +43,13 @@ end
...
 
-``` ruby
+```ruby
 The node <%= node[:fqdn] %> thinks the x-men <%= @x_men %>
 ```
 
 Would render something like:
 
-``` ruby
+```
```
tas50, line 52: This wasn't actually ruby. It was just the output of running some ruby

phiggins (approved):

IanMadd (approved):

PR #2522, opened 2020-06-08
Improve knife bootstrap docs

tas50:
I just tried to boostrap a node with the old docs. They needed some love. The example output was 8 years old and the advice was from Chef < 15.


kagarmoe (approved):
tenor-17607429

PR #2526, opened 2020-06-09
Simplify what you do from the Chef Workstation

tas50:
Talk about Policyfiles and not the legacy stuff. Call them cookbooks and avoid the what is a cookbook and what is a recipe mess.


kagarmoe (approved):
tenor-146728204

PR #2527, opened 2020-06-09
Remove some extra steps from the chef-vault bootstrap

tas50:
The gem is built into chef now. There's no need to do this.


kagarmoe (approved):
tenor-11504151

PR #2528, opened 2020-06-09
Rework the story we're telling in the validatorless bootstrap section

tas50:
When we originally wrote this the feature was new so we were letting folks know how to migrate. It's not new at this point and we warn folks not to use it when they bootstrap. Instead we should tell them *why* they shouldn't use it and show them how to move off it. I nuked the layout file that was only used in this single place. There's a ws_ version of it that's not used anywhere that should also go.


content/install_bootstrap.md
```diff
@@ -118,17 +118,17 @@ describe how to bootstrap a node using knife.
     just bootstrapped, run the following command:
 
     ``` bash
-    knife client show name_of_node
+    knife client show NAME_OF_NODE
```
kagarmoe, line 121:
```suggestion
    knife client show NODE_NAME
```

content/install_bootstrap.md
```diff
@@ -118,17 +118,17 @@ describe how to bootstrap a node using knife.
...
     ```
 
-    where `name_of_node` is the name of the node that was just
+    where `NAME_OF_NODE` is the name of the node that was just
```
kagarmoe:
```suggestion
    where `NODE_NAME` is the name of the node that was just
```

content/install_bootstrap.md
```diff
@@ -118,17 +118,17 @@ describe how to bootstrap a node using knife.
...
     bootstrapped. The Chef Infra Server will return something similar
     to:
 
     ``` bash
     admin:     false
     chef_type: client
-    name:      name_of_node
+    name:      NAME_OF_NODE
```
kagarmoe:
```suggestion
    name:      NODE_NAME
```

content/install_bootstrap.md
kagarmoe: Unless there's some reason for using the preposition?

tas50:
Thanks for the cleanup @kagarmoe

PR #2538, opened 2020-06-12
Update resources_common_properties.md

mbaitelman:
Add :quiet 
See https://github.com/chef/chef/blob/master/lib/chef/resource.rb#L540

### Description

Documents missing option 
### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/layouts/shortcodes/resources_common_properties.md
+++ b/layouts/shortcodes/resources_common_properties.md
@@ -5,7 +5,8 @@ The following properties are common to every resource:
 :   **Ruby Type:** true, false, :quiet \| **Default Value:** `false`
 
     Continue running a recipe if a resource fails for any reason.
-
+    Using `:quiet` logs the full error though will still continue running the recipe.
+    
 `retries`
 
 :   **Ruby Type:** Integer \| **Default Value:** `0`
--- a/layouts/shortcodes/resources_common_properties.md
+++ b/layouts/shortcodes/resources_common_properties.md
@@ -2,7 +2,7 @@ The following properties are common to every resource:
 
 `ignore_failure`
 
-:   **Ruby Type:** true, false \| **Default Value:** `false`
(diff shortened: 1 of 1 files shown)


layouts/shortcodes/resources_common_properties.md
```diff
@@ -2,7 +2,7 @@ The following properties are common to every resource:
 
 `ignore_failure`
 
-:   **Ruby Type:** true, false \| **Default Value:** `false`
+:   **Ruby Type:** true, false, :quiet \| **Default Value:** `false`
 
     Continue running a recipe if a resource fails for any reason.
```
tas50, line 7: We should document what quiet actually does here

layouts/shortcodes/resources_common_properties.md
```diff
@@ -2,10 +2,11 @@ The following properties are common to every resource:
...
-
+    Using `:quiet` logs the full error though will still continue running the recipe.
```
IanMadd, line 8:
```suggestion

    Using `:quiet` logs the full error and the recipe will continue to run if a resource fails.
```

IanMadd (commented):
@mbaitelman Could you add your DCO signoff.
https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

layouts/shortcodes/resources_common_properties.md
tas50: I'm not sure if that's the correct description. Using :quiet prevents a full error stacktrace from being shown.

IanMadd: So maybe

`:quiet` will not display the full stack trace and the recipe will continue to run if a resource fails.

mbaitelman:
Created #2541 to replace this PR. 
It includes the DCO and requested updates.

PR #2544, opened 2020-06-16
Add site map as a shortcode

IanMadd:

### Description

This generates a site map with contents that are nearly identical to the left nav. 

It does eliminate some external links that are in the left nav, but not all of them. Because sometimes Hugo doesn't want to do what Hugo doesn't want to do.

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


kagarmoe (approved):
tenor-113535363

PR #2547, opened 2020-06-18
Fix a note about powershell_script resources.

phiggins:
A very important correction. These two files are identical, should one get removed?

Signed-off-by: Pete Higgins <pete@peterhiggins.org>

### DIFF
--- a/layouts/shortcodes/dt_resources_common_guard_interpreter_attributes.md
+++ b/layouts/shortcodes/dt_resources_common_guard_interpreter_attributes.md
@@ -26,7 +26,7 @@ values:
 `:powershell_script`
 
 :   Evaluates a string command using the **powershell_script**
-    resource. Default value (within a **batch** resource block):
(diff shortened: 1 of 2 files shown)


IanMadd (approved):

IanMadd:
@phiggins Right now any shortcode that's used in the desktop docs has a `dt_` prefix. We have some shortcodes that are identical so that if the relevant code changes in desktop but not in infra, then we won't end up mucking things up by changing shortcodes that are used for both. (also workstation) At some point I'd like to go through the resources shortcodes and just verify that they are and will be the same regardless of repo and then we'll just establish that shortcodes that start with `resources_` can be used for both repos without messing up the docs for either product. But I haven't done that yet.

phiggins:
> We have some shortcodes that are identical so that if the relevant code changes in desktop but not in infra, then we won't end up mucking things up by changing shortcodes that are used for both. 

That makes sense. Thanks for clarifying!

PR #2548, opened 2020-06-18
Update the compile time forcing docs

lamont-granquist:
Update the docs here for Chef 16 since they were looking very
Chef 11/12-ish.

### DIFF
--- a/layouts/shortcodes/resources_common_compile_begin.md
+++ b/layouts/shortcodes/resources_common_compile_begin.md
@@ -93,7 +93,7 @@ to download the file during compile time:
 ```
 
 This is considerably cleaner than the alternative of lazying both the parsing of the
-JSON and the rendering of the data into the file template, which will be caused if
+JSON and the rendering of the data into the file template, which will happen if
 the `remote_file` resource is not run at compile time:
 
 ``` ruby
--- a/layouts/shortcodes/resources_common_compile_begin.md
+++ b/layouts/shortcodes/resources_common_compile_begin.md
@@ -20,7 +20,7 @@ end
 ```
 
 That flag both forces the resource to run at compile time and sets the
-converge action to :nothing.
+converge action to `:nothing`.
 
 The following examples show when (and when not) to use `run_action`.
 
--- a/layouts/shortcodes/resources_common_compile_begin.md
+++ b/layouts/shortcodes/resources_common_compile_begin.md
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
 the specified action during the compile phase. For example:
 
 ``` ruby
-resource_name 'foo' do
+build_essential 'Install compilers' do
   action :nothing
-end.run_action(:some_action)
+end.run_action(:install)
 ```
 
 where `action` is set to `:nothing` to ensure the `run_action` is run
 during the compile phase and not later during the execution phase.
 
+This can be simplified by using the `compile_time` flag in Chef Infra
+Client 16 and later versions:
+
+``` ruby
+build_essential 'Install compilers' do
+  compile_time true
+end
+```
+
+That flag both forces the resource to run at compile time and sets the
+converge action to :nothing.
+
 The following examples show when (and when not) to use `run_action`.
 
-**Update a package cache**
+**Using Custom Resources preferred to forcing to compile time**
 
-Sometimes it is necessary to ensure that an operating system's package
-cache is up to date before installing packages. For example, on Debian
-or Ubuntu systems, the Apt cache should be updated:
+Compile time execution is often used to install gems before requiring
+them in recipe code.
+
+This is a poor pattern, since gems may depend on native gems, which
+may result in needing to install compilers at compile time.
 
 ``` ruby
-if node['apt']['compile_time_update'] && ( !::File.exist?('/var/lib/apt/periodic/update-success-stamp') || !::File.exist?(first_run_file) )
-  e = bash 'apt-get-update at compile time' do
-    code <<-EOH
-      apt-get update
-      touch #{first_run_file}
-    EOH
-    ignore_failure true
-    only_if { apt_installed? }
-    action :nothing
+build_essential 'Install compilers' do
+  compile_time true
+end
+
+chef_gem 'aws-dsk' do
+  compile_time true
+end
+
+require 'aws-sdk'
+```
+
+A better strategy is to move the code which requires the gem into
+a custom resource.  Since all the actions of custom resources run
+at converge time, this has the effect of deferring the require of
+the gem to later in the overall chef-client execution.  Unified
+mode can also be used in the resource to eliminate compile/converge
+mode issues entirely:
+
+``` ruby
+unified_mode true
+
+action :run do
+  build_essential 'Install compilers'
+
+  chef_gem 'aws-sdk'
+
+  require 'aws-sdk'
+end
+```
+
+**Download and parse a configuration file**
+
+A common use case is to download a configuration file, parse it and then
+use the values in templates and to control other configuration.
+
+And important distinction to make is that the downloaded configuration file
+is being used internally by the Chef Infra Client as temporary state.
+
+To download and parse a JSON file and render it in a template it makes sense
+to eagerly download the file in compile time:
+
+``` ruby
+  # the remote_file is being downloaded to a temporary file
+  remote_file "#{Chef::Config[:file_cache_path]}/users.json" do
+    source "https://jsonplaceholder.typicode.com/users"
+    compile_time true
   end
-  e.run_action(:run)
-end
+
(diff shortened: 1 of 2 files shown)


layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+**Using Custom Resources preferred to forcing to compile time**
+
+Compile time execution is often used to install gems before requiring
+them in recipe code.
 
-Sometimes it is necessary to ensure that an operating system's package
-cache is up to date before installing packages. For example, on Debian
-or Ubuntu systems, the Apt cache should be updated:
+This is a poor pattern, since gems may depend on native gems, which
+may result in needing to install compilers at compile time.
```
IanMadd:
```suggestion
This is a poor pattern since gems may depend on native gems which
may require installing compilers at compile time.
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+require 'aws-sdk'
+```
 
-A very common use case is to install a gem during the compile phase so
-that it will be available to Chef Infra Client during the execution
-phase. This is why the **chef_gem** resource exists. For example, this:
+A better strategy is to move the code which requires the gem into
+a custom resource.  Since all the actions of custom resources run
+at converge time, this has the effect of deferring the require of
+the gem to later in the overall chef-client execution.  Unified
```
IanMadd:
```suggestion
A better strategy is to move the code, which requires the gem, into
a custom resource. Since all the actions of custom resources run
at converge time, this defers requiring 
the gem until later in the overall Chef Infra Client execution. Unified
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+  chef_gem 'aws-sdk'
+
+  require 'aws-sdk'
 end
 ```
 
-is effectively the same as
+**Download and parse a configuration file**
+
+A common use case is to download a configuration file, parse it and then
```
IanMadd:
```suggestion
A common use case is to download a configuration file, parse it, and then
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+use the values in templates and to control other configuration.
+
+And important distinction to make is that the downloaded configuration file
+is being used internally by the Chef Infra Client as temporary state.
+
+To download and parse a JSON file and render it in a template it makes sense
```
IanMadd:
```suggestion
To download and parse a JSON file and render it in a template, it makes sense
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+to eagerly download the file in compile time:
```
IanMadd:
```suggestion
to download the file during compile time:
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
 ```
 
-is effectively the same as
+**Download and parse a configuration file**
+
+A common use case is to download a configuration file, parse it and then
+use the values in templates and to control other configuration.
+
+And important distinction to make is that the downloaded configuration file
+is being used internally by the Chef Infra Client as temporary state.
```
IanMadd: Not sure of the context here but maybe one of these suggestions: 

```suggestion
An important distinction to make is that the downloaded configuration file
only exists in a temporary state to be used by the Chef Infra Client. It will
not be used directly by the system or applications that are managed by the
Chef Infra Client.
```
Or maybe:

`used internally while the Chef Infra Client is in a temporary state.`

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+  array = JSON.parse(IO.read("#{Chef::Config[:file_cache_path]}/users.json")
+
+  # the `array.last["phone"]` expression here will also be evaluated at compile
+  # time and must be lazied via wrapping the expresssion in `lazy {}`
+  file "/tmp/phone_number.txt" do
+    content array.last["phone"]
+  end
+```
+
+This is considerably cleaner than the alternative of lazying both the parsing of the
```
IanMadd:
```suggestion
This is considerably cleaner than the alternative of lazy evaluating both the parsing of the
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+JSON and the rendering of the data into the file template, which will be caused if
```
IanMadd:
```suggestion
JSON and the rendering of the data into the file template, which will happen if
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+  end
+
+  # the argument to the content property must now also be deferred
+  file "/tmp/phone_number.txt" do
+    content lazy { array.last["phone"] }
+  end
 ```
 
-but without needing to define a `run_action`.
+This is an example of code that overly uses deferred execution, uses more "lazy" evaluation, and is
```
IanMadd:
```suggestion
This is an example of code that overuses deferred execution, uses more "lazy" evaluation, and is
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -70,13 +129,28 @@ resources. For example:
...
\ No newline at end of file
+to other resources later on.
+
+The best approach to this problem is to use `unified mode` which eliminates
+the compile time and converge time distinction, while allowing notifications
+to work correctly.
+
+**Resources that are forced to compile time by default**
+
+The `ohai_hint` and `hostname` resources are forced to compile time by default.
```
IanMadd:
```suggestion
The `ohai_hint` and `hostname` resources run at compile time by default.
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -70,13 +129,28 @@ resources. For example:
...
+
+This is due to the fact that later resources may consume the node attributes which
+are set by those resources leading to excessive use of `lazy` in subsequent
+resources (and similar issues to the `remote_file` example above).
+
+The `chef_gem` resource used to be forced to compile time by default, but that was
+removed in favor of the recommendation to move such code to custom resources.
```
IanMadd:
```suggestion
The `chef_gem` resource used to execute at compile time by default, but now we
recommend that users move code that executes at compile time to custom resources.
```

layouts/shortcodes/resources_common_compile_begin.md
```diff
@@ -2,65 +2,124 @@ Use `.run_action(:some_action)` at the end of a resource block to run
...
+Client 16 and later versions:
+
+``` ruby
+build_essential 'Install compilers' do
+  compile_time true
+end
+```
+
+That flag both forces the resource to run at compile time and sets the
+converge action to :nothing.
```
IanMadd:
```suggestion
converge action to `:nothing`.
```

IanMadd (commented):
Some suggestions for clarity.

layouts/shortcodes/resources_common_compile_begin.md
lamont-granquist: definitely not the second one.

lamont-granquist: So the file that is being downloaded does not really affect the running system or the applications.  It isn't a file in /etc, it isn't some kind of O/S config, it isn't an application config file.  So the file is downloaded, then the recipe reads it, then the chef-client goes and does the REAL work of configuring the system.  So it is internal or temporary or transient state that is only consumed by chef-client itself.

Dunno if that helps.

Also, the file clearly gets downloaded to the filesystem so technically the running state of the node does change.  But the difference I'm trying to highlight is that the change in the state is entirely to a "scratchpad", or that the downloaded file is part of the "preparation" to do actual work.

IanMadd (approved):

PR #2558, opened 2020-06-25
Change the link to more information on ssl_ciphers

btm:
The mozilla page doesn't contain useful information on the format of
ssl_ciphers, which is actually an OpenSSL configuration value. The
OpenSSL documentation is much more useful.


IanMadd (approved):
tenor-223382066

PR #2575, opened 2020-07-09
Remove Debian 8, include Debian 9/10 with Workstation

mattray:
8 was EOL'd https://www.debian.org/releases/jessie/
@tas50 said to add Debian to Workstation

Signed-off-by: Matt Ray <github@mattray.dev>


tas50 (approved):

tas50:
We're missing some other stuff we support now, but this is better

PR #2579, opened 2020-07-17
Remove language about default value for path.

gyamada619:
### Description

The `path` property no longer has the default value of the resource block's name; that belongs to `source`.

### Definition of Done

Remove text as specified.

### Issues Resolved

None.

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


tas50:
Not sure how that came in. Thanks for the fix @gyamada619

PR #2586, opened 2020-07-23
Revert "Remove Automate v1 requirements"

tas50:
This is going to seem a bit backwards, but this was accidentally committed to mater and I wanted to make sure everyone was OK with the change. If everyone is fine with the change we can delete this PR. If not we can merge the revert.


mjingle:
Thanks for letting us know about this, Tim. We got confirmation that we do not need to revert this merge.

PR #2616, opened 2020-08-20
date command needs a format for epoch timestamps

mattray:
if you don't have this, you'll end up with strings in your dates and hab doesn't like that


IanMadd:
Hi Matt, can you make this change in https://github.com/chef/effortless/tree/master/effortless-chef-io/content/effortless

PR #2618, opened 2020-08-21
install_bootstrap unattended script updates

collinmcneese:
### Description

Changes the following for `install_bootstrap.md`
- updates  powershell code block to type `powershell` for visual rendering.
- updates unattended bash `client.rb` section to use heredoc rather than `echo -e`.  The `echo -e` method causes formatting issues with the `client.rb` file that is rendered with an out-of-the-box MacOS installation, using heredoc works with both MacOS and Linux.
- removes `sudo` from `chef-client` execution at the end of the unattended bootstrap file, this is not required in this context as the entire script must be run with `sudo`

### Check List

- [X] Spell Check
- [X] Local build
- [X] Examine the local build
- [] All tests pass

### DIFF
--- a/content/install_bootstrap.md
+++ b/content/install_bootstrap.md
@@ -429,7 +429,7 @@ bootstrapping Windows and Linux nodes.
 
 #### PowerShell User Data
 
-``` none
+``` powershell
 ## Set host file so the instance knows where to find chef-server
 $hosts = "1.2.3.4 hello.example.com"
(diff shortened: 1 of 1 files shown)


phiggins (approved):
Makes sense to me. 👍

IanMadd (approved):

IanMadd:
chef/release-engineering#1259

PR #2653, opened 2020-09-15
Plan tabs using foundation

kagarmoe:

### Description

Shortcode for presenting two code examples that are always in `bash` and `powershell`

It needs to:
#. Take two fenced code examples
#. And present them inside of two tabs with the correct code highlighting

so:

"```bash plan.sh
...some code
```

"```ps plan.ps1
...some code
```""

Are presented as:
 // Working on this


IanMadd:
See #2659

Issue #2665, opened 2020-09-23
Chef-solo documentation may cause security issues on implementations


bennyvasquez:
(submitted on behalf of user [from discourse](https://discourse.chef.io/t/chef-solo-documentation-may-cause-security-issues-on-implementations/17718))

## Description
I'd like to inform you about one potential risk where system administrators might be misled to where the "chef" user is not supposed to have root privileges on the host system.

[chef-solo (executable)](https://docs.chef.io/ctl_chef_solo/#run-as-non-root-user)

Since, non-root users can run Ruby scripts with passwordless sudo privileges defined in the sudoers file then the "chef" user can create a Ruby file containing "system('/bin/bash')" in order to get a root shell.

```
chef@localhost$ sudo chef-solo -c /tmp/shell.rb
root@localhost#
```

Therefore, I think the documentation might need some warning or definition where system administrators need to use chef-solo accessible for unprivileged users who should not become the built-in root user.

Kind regards.

damacus:
Resolved by #2667

PR #2667, opened 2020-09-23
Added a security warning under the sudo configuration on chef-solo markdown

g0ku704:
### Description

The guideline for sudo configuration on chef-solo may cause security issues depeding of the implementation of chef-solo in the infrastructre.
As it has been described in the issue, the risk of privilege escalation might happen in cases where 'chef' user should not be able to have built-in root rights on the system.

### Definition of Done

A warning message added for users to review their implementation in order to use sudo with non-root user for chef-solo executable.

### Issues Resolved

https://github.com/chef/chef-web-docs/issues/2665

### Check List

- [x] Spell Check
- [x] Local build
- [x] Examine the local build
- [x] All tests pass


IanMadd:
Hi @rustempasha, thanks for your PR. Can you add your DCO signoff to this PR. 
See here --> https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

content/ctl_chef_solo.md
```diff
@@ -190,6 +190,19 @@ chef ALL=(ALL) NOPASSWD: /usr/bin/chef-solo
...
+
+The configuration for the 'chef' user as described above will still 
+going to have root-level access by calling system commands from a Chef 
+script file with sudo privileges. 
+
+To prevent such kind of security issues, it is highly recommended to create 
+an alternate approach for the cases where 'chef' user should not have 
+built-in root level access.
+
+{{< warning >}}
```
phiggins:
```suggestion
{{< /warning >}}
```

This needs to be a closing tag, otherwise you're opening a second warning tag within the first. 

Screen Shot 2020-09-23 at 10 57 09 AM

content/ctl_chef_solo.md
```diff
@@ -190,6 +190,19 @@ chef ALL=(ALL) NOPASSWD: /usr/bin/chef-solo
...
 
+{{< warning >}}
+
+The configuration for the 'chef' user as described above will still 
+going to have root-level access by calling system commands from a Chef 
+script file with sudo privileges. 
+
+To prevent such kind of security issues, it is highly recommended to create 
+an alternate approach for the cases where 'chef' user should not have 
+built-in root level access.
```
phiggins:
```suggestion
Following the instructions above will effectively grant the `chef` user root-level access to system commands by allowing them to run arbitrary Ruby code in the form of Chef scripts with sudo privileges. To prevent security issues, it is highly recommended that this is only done on systems with trusted users.
```

I rewrote this with a few thoughts in mind:

* I tried to make the text more explicit about what the possible consequences are.
* Mentioning "an alternate approach" without saying what that might be feels incomplete.

Does this enable a certain use case, or make it more convenient? It might be worth adding information about why the user would want to do this so that users are not led to believe they need to do this.

g0ku704: I think people implement such configuration might think to provide an option to run Chef scripts locally. To have a better approach can be proposed is to allow only to run "sudo chef-solo" with read-only Chef scripts where located outside of 'chef' user's home folder like
chef ALL=(ALL) NOPASSWD: /usr/bin/chef-solo -c /lib/scripts/chef.rb (where chef user only have read permission on this file)
Although, /lib/scripts/chef.rb should not be able to run any other set of commands where "chef" user can execute.

kagarmoe (approved):
tenor-1812412

PR #2699, opened 2020-10-05
Remove uninstall from the menu

tas50:
This shouldn't be in our getting started guide. We already link to it
from the troubleshooting guide and it's still in the search.


kagarmoe:
HI @tas50! This one is just a question of manners. We told them how to get it onto their system, so we're also going to tell them how to get it back off, without making them dig around in the docs to find it. This is the once in a blue moon where I'm not going to accept your PR.

tas50:
It feels fair to not start the process out with the uninstall instructions and in general we're better documenting and promoting the uninstall process than the install process. It feels really strange to me. Similar products don't include uninstall docs when you just use the system package manager to uninstall them.

kagarmoe:
It is a topic that we can revisit when we re-org the nav.

PR #2719, opened 2020-10-22
Code Block Formatting

kagarmoe:
Makes code blocks readable by pre-wrapping. Reduces the spacing between paragraphs and code blocks by setting a negative value on the button.copy-code-button.


IanMadd (approved):
tenor-105959780

PR #2720, opened 2020-10-22
List formatting

kagarmoe:

Ideas for better lists.


IanMadd:
This one's a bit weird. This set of list items are bolded: 
https://deploy-preview-2720--chef-web-docs.netlify.app/automate/install/#backup-externally-deployed-elasticsearch-to-aws-s3

and these aren't because Hugo didn't put `<p>` inside each list item (the MD looks the same though 🤷) :
https://deploy-preview-2720--chef-web-docs.netlify.app/automate/install/#backup-externally-deployed-elasticsearch-to-local-filesystem

IanMadd:
Another thing to consider is the 1st example here, which would style the ordered list numbers: https://css-tricks.com/custom-list-number-styling/

kagarmoe:
@IanMadd This one uses bold numbers. I'm on the fence if it helpful or distracting.

IanMadd:
It doesn't really seem to be changing that much.

I'm comparing this: 
https://deploy-preview-2720--chef-web-docs.netlify.app/automate/datafeed/#edit-a-data-feed-instance

to this: 
https://docs.chef.io/automate/datafeed/#edit-a-data-feed-instance

My general instinct is that if the change doesn't really improve the readability, then it just becomes code that has to be maintained later so we shouldn't make the change.

kagarmoe:
I'll return to this when I have more time.

PR #2740, opened 2020-11-05
Consolidate two resource parameters into multi_package_resource param

### LINKED ISSUE #2731: Combine `common_resource_functionality_multiple_packages` and `properties_multiple_packages`
The `common_resource_functionality_multiple_packages` and `properties_multiple_packages` configs in the chef infra resources yaml configs should be combined into a single `multi_package_resource` config. We never use just one of these configs and that means we have to manage both in the config automation.

IanMadd:

### Description

Consolidates two resource parameters, `common_resource_functionality_multiple_packages` and `properties_multiple_packages`, into one parameter, `multi_package_resource`.

Double check that the **package** resource is correct. It had the `properties_multiple_packages` param but not the `common_resource_functionality_multiple_packages` param.

### Issues Resolved

#2731

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/resources/apt_package/_index.md
+++ b/content/resources/apt_package/_index.md
@@ -1,7 +1,6 @@
 ---
 resource_reference: true
-common_resource_functionality_multiple_packages: true
-properties_multiple_packages: true
+multi_package_resource: true
 resources_common_guards: true
 resources_common_notification: true
 resources_common_properties: true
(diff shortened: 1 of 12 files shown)


kagarmoe (approved):
Thanks!

PR #2749, opened 2020-11-12
Pull in an exact version of the docs at project promotion time

schisamo:
This change ensures we are vendoring in the exact version that was promoted of the documentation source....instead of what the latest on the default branch of the project.


IanMadd (approved):
LGTM

tduffield (approved):

PR #2754, opened 2020-11-20
Google Analytics WIP

IanMadd:
### Description

This does a few things:

- Adds Google Analytics to the head of each page.
- The Google Analytics component will only work during production deploys

- Adds feedback buttons to bottom of each page. 
  - User can click on yes or no feedback buttons
  - After a user clicks on a button:
     - the clicked button displays a thank you message
     - both buttons are disabled until the page is reloaded
     - Feedback is sent to GA with the ga() javascript function which includes data about
        - the page that was clicked on
        - binary 0 or 1 if it page was helpful (1) or not (0)
        - a few other values 

I'm not sure about the specifics of the ga() function. I don't know if it's loaded with the `_internal/google_analytics.html` template, or if there are other data we want to include, or data we want to exclude or rename.

Also, it might not be added yet to the some of the weird templates like resources or runbook pages. I haven't checked those yet.

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


kagarmoe:
Thanks for thinking about disabling the buttons after they're used.

mjingle (approved):
tenor-30412870

kagarmoe:
Switched it to "ready" and I'm going to go ahead and approve & merge. I won't know if the GA side works until this is live.

PR #2756, opened 2020-11-23
Update platforms support to EOL macos 10.13 & add a macOS policy

tas50:
We never actually had a policy defined for macOS either. Apple has no official policy, but in general they support N-2 releases and then silently stop shipping security releases for the previous releases. We have historically considered the date of the new release to be the trigger for the EOL, even if another security update may still ship.


content/platforms.md
```diff
@@ -631,8 +631,8 @@ according to those vendors' terms:
 </thead>
 <tbody>
 <tr class="odd">
-<td>AIX</td>
-<td>IBM End of Support Date</td>
+<td>Apple macOS</td>
+<td>Apple has no official EOL dates published, but generally supports N-2 macOS releases</td>
```
kagarmoe:
```suggestion
<td>Apple generally supports the last three macOS releases, for example: 10.14, 10.15, and 11.0. Apple does not officially publish EOL dates.</td>
```

I think we should avoid making customers do macOS version release algebra.

kagarmoe (changes requested):
One minor change.

kagarmoe (approved):
tenor-161131390

PR #2759, opened 2020-11-25
Use GA UA and gtags

kagarmoe:
### Description

We're not getting the full range of analytics that I want to see. This tries to add the UA and G tags. 

Based on:
https://developers.google.com/analytics/devguides/collection/ga4/basic-tag?technology=gtagjs
https://github.com/gohugoio/gohugoioTheme/blob/master/layouts/partials/gtag.html

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


mjingle (approved):
tenor-186784892

config.toml
```diff
@@ -80,12 +80,14 @@ notAlternative = true
...
   enable_search = true
   robots = ''
+  images = ["chef-progress-logo.svg"]
   menuOrder = ["overview", "automate", "compliance",  "desktop",  "infra", "habitat", "inspec", "workstation", "effortless", "release_notes", "legacy", "extra"]
 
 [taxonomies]
   category = "categories"
   tag = "tags"
   resource = "resources"
+  series = "series"
```
IanMadd: Do we need these changes?

IanMadd:
One thing that might make things simpler would be to define the two different tags in the config.toml, 

```
googleAnalyticsUA = "UA-183953607-1"
googleAnalyticsG = "G-CH0D7HCR73"
```

And then just call them: `{{ $.Site.Params.googleAnalyticsUA }}` or `{{ $.Site.Params.googleAnalyticsG }}`.

Something like that.

themes/docs-new/layouts/partials/gtag.html
```diff
@@ -0,0 +1,27 @@
+{{ with .Site.GoogleAnalytics }}
```
IanMadd:
```suggestion
{{ with .Site.GoogleAnalytics }}
```
We can ditch the `{{ with .Site.GoogleAnalytics }}`, and also the `{{end}}` at the bottom.

config.toml
kagarmoe: Not in this PR...it is noise from something else I'm working on.

IanMadd:
In the gtag partial there's this bit: `'dimension2': '{{ getenv "BRANCH" }}`. I think Hugo has that in their theme because they're doing deploys from two branches so I think they're tracking data from both. I could be wrong about that, but I don't think we'll get anything out of that. But it also won't hurt.

IanMadd (approved):

kagarmoe:
I think it might be helpful for A/B testing, although I'm not sure.

IanMadd:
We might want to remove the other part too `'dimension1': '{{ $site }}'`. Hugo uses that theme for two sites so they're tracking both.

themes/docs-new/layouts/partials/gtag.html
```diff
@@ -0,0 +1,26 @@
+<!-- Global site tag (gtag.js) - Google Analytics -->
+<script async src="https://www.googletagmanager.com/gtag/js?id={{ $.Site.Params.googleAnalyticsUA }}"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
+  gtag('js', new Date());
+
+  {{ $site := $.Site.BaseURL | replaceRE "^https?://(www\\.)?([^/]+).*" "$2" }}
```
IanMadd: We can cut that out too.
```suggestion
```

PR #2765, opened 2020-12-01
Misc cleanup to the attributes documentation

tas50:
Not great, but better.


content/attributes.md
```diff
@@ -867,3 +806,21 @@ For adding a string using a multi-level array:
     role_or_environment 2 [ [ 3 ] ]
     =
     [ [ 1, 2 ], [ 3 ] ]
+
+## Limiting Attribute Persistence
+
+Some organizations find it helpful to control attribute data stored to the Chef Infra Server in order to limit the disk and CPU resources used processing unused attributes. For example your organization may find the data from the Ohai `Package` plugin useful when writing cookbooks, but you don't see the need in savings ~100k of package information for each Chef Infra Client run. By limiting this data from being saved to the Chef Infra Server it will still be available on the node within cookbooks, but won't be saved to the Chef Infra Server, where it is available in searches.
```
IanMadd:
```suggestion
Some organizations find it helpful to control attribute data stored to the Chef Infra Server in order to limit the disk and CPU resources used when processing unused attributes. For example, your organization may find the data from the Ohai `Package` plugin useful when writing cookbooks, but you don't see the need in saving ~100k of package information for each Chef Infra Client run. By limiting the data that is saved to the Chef Infra Server, it will still be available on the node within cookbooks, but won't be saved to the Chef Infra Server where it is available in searches.
```

themes/docs-new/layouts/shortcodes/ohai_automatic_attribute.md
```diff
@@ -21,38 +21,42 @@ Infra Client run. The most commonly accessed automatic attributes are:
 <td>The platform on which a node is running. This attribute helps determine which providers will be used.</td>
 </tr>
 <tr class="even">
+<td><code>node['platform_family']</code></td>
+<td>The platform family is a Chef Infra specific grouping of similar platforms where cookbook code can often be shared. For example `rhel` includes Red Hat Linux, Oracle Linux, CentOS, and several other platforms that are nearly identical to Red Hat Linux.</td>
```
IanMadd:
```suggestion
<td>The platform family is a Chef Infra specific grouping of similar platforms where cookbook code can often be shared. For example, `rhel` includes Red Hat Linux, Oracle Linux, CentOS, and several other platforms that are nearly identical to Red Hat Linux.</td>
```

IanMadd (approved):

tas50:
Just a heads up this needs a lot of other things to be updated. I'm trying to figure out when I can take the additional time to update everything else.

IanMadd:
@tas50 Does it make sense to merge this and then you can come back and edit other things when you have time?

tas50:
There's a few more things that are needed for this to make sense. From there we can iterate on the next phase.

content/attributes.md
```diff
@@ -68,43 +68,6 @@ Infra Client chooses which attribute to apply.
 </tbody>
 </table>
 
-## Attribute Sources
```
kagarmoe, line 71: What's your reasoning for removing this information? Understanding where attributes are defined is important.

content/attributes.md
```diff
@@ -117,9 +80,8 @@ An attribute file is located in the `attributes/` sub-directory for a
 cookbook. When a cookbook is run against a node, the attributes
 contained in all attribute files are evaluated in the context of the
 node object. Node methods (when present) are used to set attribute
```
kagarmoe: We need to define the term "node methods"

content/attributes.md
```diff
@@ -671,10 +613,7 @@ node['foo']['bar'] #=> {'baz' => 99}
 
 ## About Deep Merge
 
-Attributes are typically defined in cookbooks, recipes, roles, and
-environments. These attributes are rolled-up to the node level during a
-Chef Infra Client run. A recipe can store attribute values using a
-multi-level hash or array.
+Attributes are typically defined in cookbooks, recipes, and Policyfiles. These attributes are rolled-up to the node level during a Chef Infra Client run. A recipe can store attribute values using a multi-level hash or array.
```
kagarmoe: We need to define the term "deep merge" or not use it at all. Is "deep merge" the process that Chef Infra Client uses to negotiate nuanced attribute definitions from a variety of sources during a run? Something else?

"are typically defined": Is this a best practice (or not)? Do we want users to define attributes in cookbooks, recipes, and Policyfiles and avoid other ways? 

"rolled-up" is idiomatic. "The Chef Infra Client applies attributes defined in these sources to the node during a run." or something similar.

content/attributes.md
```diff
@@ -867,3 +806,21 @@ For adding a string using a multi-level array:
...
+
+{{< note >}}
+
+In Chef Infra Client 16.3 the node Blacklist and Whitelist features were renamed to Blocklist and Allowlist. For backwards compatibility the old configuration values will continue to work, but this document will describe the Blocklist and Allowlist names. See each section below for the appropriate legacy configuration values if you are running legacy clients in your organization.
+
+{{< /note >}}
+
+### Attributes Blocklist
+
+{{% node_attribute_blacklist %}}
```
kagarmoe: Let's rename the files, while we're at it.

themes/docs-new/layouts/shortcodes/node_attribute.md
```diff
@@ -1,35 +1,22 @@
...
 
 Attributes are defined by:
 
--   The state of the node itself
--   Attributes passed via JSON on the CLI
--   Cookbooks (in attribute files and/or recipes)
--   Roles
--   Environments
--   Policyfiles
+- The node as saved on the Chef Infra Server
```
kagarmoe, line 9: Maybe I don't understand, but this seems circular to me. It is like saying a node sets its own attributes. My understanding is that node that hasn't had a Chef Infra Client run has no attributes, and it is only after a run that the node object with default attributes exists for Chef Infra Server to store. The Chef Infra Client run uses the node object as an index, against which it compares existing node content against content that it needs to download. Beyond comparing the list of files that already exist on the server, the Chef Infra Client removes and then rewrites the node object. Or am I missing something?

kagarmoe (changes requested):
Many comments and a few places that we need to change.

In a very real way, I'm not sure what this document is attempting to achieve. It combines concept, reference, and tutorial pieces. It is alternately formal and informal.

We need to decide the audience(s) and purpose(s) of this content before we can fix it.

content/attributes.md
tas50: Yep I have to rewrite them entirely as well

kagarmoe (approved):

content/attribute_precedence.md
```diff
@@ -0,0 +1,481 @@
...
+
+```ruby
+default_attributes({ 'apache' => {'dir' => '/etc/apache2'}})
+```
+
+**Normal attribute set as a cookbook attribute**
+
+```ruby
+set['apache']['dir'] = '/etc/apache2'
+normal['apache']['dir'] = '/etc/apache2'  #set is an alias of normal.
```
lamont-granquist: node.set was removed after chef-12 so it doesn't work chef >= 13

lamont-granquist (approved):

PR #2766, opened 2020-12-01
Create a pipeline for triggering during head of channel promotions

schisamo:
For projects that use the "head of channel" promotion pattern (e.g. Automate and Habitat) we cannot leverage subscriptions configured in this repo as many of the promotions tasks that are executed in said promotions will not be complete when our bash script is executed (think updating project manifests).

In order to work around this we will create a new pipeline which can be triggered (using Expeditor's `trigger_pipeline`) as one of the actions in the `project_promoted` workload subscription in the actual projects being promoted. Although it's not ideal that our projects which use the "head of channel" promotion pattern will need to know about `chef/chef-web-docs` this approach avoids nasty workaround like `sleep` statements in our bash script.

### DIFF
--- a/.expeditor/buildkite/update_hugo_modules_project_promoted.sh
+++ b/.expeditor/buildkite/update_hugo_modules_project_promoted.sh
@@ -2,21 +2,30 @@
 
 set -eoux pipefail
 
+# Both of these values should be set when this script is executed in the
+# `head_of_channel_project_promoted` pipeline and triggered via the
+# `trigger_pipeline` built-in action. See more at:
+#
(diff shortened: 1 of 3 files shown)


nkierpiec (approved):
Yeah no real elegant way to do this at the moment with cross project subscriptions that trigger on promotions. 

Could potentially look into some webhook that connects into package router and triggers an event off of a newly generated manifest or checking time stamps in the bash scripts but you said you wanted to avoid sleeps.

Other than that, yeah it looks fine to me.

schisamo:
Closing this in favor of https://github.com/chef/chef-web-docs/pull/2767

PR #2768, opened 2020-12-02
Add missing platform_family attribute

tas50:

### DIFF
--- a/themes/docs-new/layouts/shortcodes/ohai_automatic_attribute.md
+++ b/themes/docs-new/layouts/shortcodes/ohai_automatic_attribute.md
@@ -21,38 +21,42 @@ Infra Client run. The most commonly accessed automatic attributes are:
 <td>The platform on which a node is running. This attribute helps determine which providers will be used.</td>
 </tr>
 <tr class="even">
+<td><code>node['platform_family']</code></td>
+<td>The platform family is a Chef Infra specific grouping of similar platforms where cookbook code can often be shared. For example `rhel` includes Red Hat Linux, Oracle Linux, CentOS, and several other platforms that are nearly identical to Red Hat Linux.</td>
+</tr>
+<tr class="odd">
 <td><code>node['platform_version']</code></td>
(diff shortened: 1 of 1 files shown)


IanMadd (approved):

themes/docs-new/layouts/shortcodes/ohai_automatic_attribute.md
```diff
@@ -21,38 +21,42 @@ Infra Client run. The most commonly accessed automatic attributes are:
 <td>The platform on which a node is running. This attribute helps determine which providers will be used.</td>
 </tr>
 <tr class="even">
+<td><code>node['platform_family']</code></td>
+<td>The platform family is a Chef Infra specific grouping of similar platforms where cookbook code can often be shared. For example `rhel` includes Red Hat Linux, Oracle Linux, CentOS, and several other platforms that are nearly identical to Red Hat Linux.</td>
```
kagarmoe, line 25:
```suggestion
<td>`platform_family` is a Chef Infra attribute that groups together similar platforms in order to support sharing cookbook code. For example `[platform_family][rhel]` includes Red Hat Linux, Oracle Linux, CentOS, and several other platforms that are nearly identical to Red Hat Linux.</td>
```

kagarmoe (changes requested):

themes/docs-new/layouts/shortcodes/ohai_automatic_attribute.md
tas50: InSpec doesn't actually use this value. In a perfect world that would be true, but this is 100% an infra thing.

kagarmoe: OmO. It calls `platform_family` to match `os.family`.

tas50: os.family and platform_family do not match.  It's a mess, but we're stuck with it unless one of you has a time machine.

kagarmoe (approved):

PR #2770, opened 2020-12-02
Document the replacement of the blacklist / whitelist in Infra Client

tas50:
We renamed these configs and it never made it to the docs. We also don't explain why you'd want to use the feature very well.


kagarmoe (approved):
tenor-224258649

PR #2776, opened 2020-12-02
Don't make policyfiles seem optional

tas50:
It's not 2016 anymore.


content/policyfile.md
```diff
@@ -16,21 +16,17 @@ aliases = ["/policyfile.html"]
...
--   Focus the workflow on the entire system
--   Safer development workflows
--   Less expensive computation
--   Code visibility
--   Role mutability
--   Cookbook mutability
--   Replaces Berkshelf and the environment cookbook pattern
+## Why Policyfiles?
+
+Policyfiles make it easier to test and promote code safely with a simpler interface. Policyfile improves the user eexperience and resolvs real-world problems that some workflows built around Chef Infra must deal with. The following sections discuss in more detail some of the good reasons to use Policyfile, including:
```
phiggins:
```suggestion
Policyfiles make it easier to test and promote code safely with a simpler interface. Using a Policyfile improves the user experience and resolves real-world problems that some workflows built around Chef Infra must deal with. The following sections discuss in more detail some of the good reasons to use Policyfile, including:
```

content/policyfile.md
```diff
@@ -60,17 +56,14 @@ cookbook or role from being applied immediately to all machines.
 
 ### Code Visibility
 
-When running Chef without Policyfile, the exact set of cookbooks that
+When running Chef Infra without Policyfile, the exact set of cookbooks that
```
phiggins:
```suggestion
When running Chef Infra without a Policyfile, the exact set of cookbooks that
```

There are a few examples like this and it's hard to tell if it's intentional. Is it meant to be used without an article like this?

IanMadd (approved):

PR #2780, opened 2020-12-03
Move two generated habitat files to content

IanMadd:

### Description

Every time you run `hugo mod vendor`, Hugo deletes everything in `_vendor` and grabs all of the vendored content from the source repos all over again. So these vendored files that are overwritten by Expeditor will just revert back to the state that they're in in the Hab repo whenever we update the vendored docs for any of the other repos.

The solution is to just dump those two files into `content`.

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


schisamo (approved):

kagarmoe (approved):
tenor-190245086

PR #2790, opened 2020-12-14
Prep Push Jobs for EOL

kagarmoe:

### Description

Push Jobs is EOL December 31, 2020. This PR moves the main content into the "Legacy" section.
- Reduces server content
- Simplifies EOL process (We will still need to EOL partials and line-edits.

### Issues Resolved

### Check List

- [ ] Spell Check
- [x] Local build
- [ ] Examine the local build
- [ ] All tests pass


IanMadd (approved):
My only suggestion would be to change the menu weight for either the Push Jobs or Workflow section so they don't have the same value. Otherwise, lgtm.

Issue #2775, opened 2020-12-02
Edit on Github


IanMadd:
### User Story
Find a way to create Edit on GitHub links using Hugo so it's automated and not a manual link at the top of each markdown page.
The big issue is getting the links to link to the md files in their original repos and not to the vendored files.

Repos:
- [X] chef-web-docs
- [X] chef-server
- [x] chef-workstation
- [x] automate
- [x] desktop-config
- [x] effortless
- [x] inspec
- [x] habitat

![tenor-38409358](https://user-images.githubusercontent.com/8138169/100925247-9a18fe00-3496-11eb-8751-f16a2b769ad3.gif)

IanMadd:
https://github.com/google/docsy/blob/master/layouts/partials/page-meta-links.html

IanMadd:
Done

PR #2798, opened 2020-12-18
Create partial for links to github pages

IanMadd:

### Description

This replaces "Edit on GitHub" links in the page text with a partial template.
This only works on pages with a `gh_repo` parameter and pages that are generated from files, so pages from other repos aren't affected until that parameter is set. Pages that shouldn't have that parameter (streamlines release notes, resource pages, runbook page) aren't affected. 

I've also added it to the versioned docs pages. 

The partial gets the value of the `gh_repo` page parameter, then gets the matching site level parameter for the repository which is set in the `config.toml`, and returns the path specified by that parameter. Then it adds the `.File.Path` variable for each page and makes a link.

#2775

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


kagarmoe (approved):
tenor-229990072

PR #2801, opened 2020-12-21
Move Chef Manage to Legacy

kagarmoe:

### Description

Preparing the way for EOLing Manage on December 31, 2021

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -560,19 +560,6 @@ identifier = "chef_infra"
     parent = "chef_infra/features"
     weight = 20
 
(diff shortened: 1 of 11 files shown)


tas50 (approved):
This feels like the right thing in 2021. Legacy means it's deprecated not necessarily EOL. We want to signal to new users that this is not the path forward.

PR #2809, opened 2020-12-23
Update the server upgrade page for EOY EOLs

tas50:
Rework the version table for clarify
Remove Push jobs as it goes EOL at the EOY
Update Chef Server 12 to be considered EOL since it also goes EOL at the end of the year
Clarify that Chef Infra Server 13 requires EULA acceptance and link to the right place


kagarmoe:
Closing because we're moving the docs! Please reopen in chef/chef-server.

PR #2818, opened 2020-12-30
Remove a mention of ChefDK

tas50:
Found another place we needed to remove DK in the docs


kagarmoe (approved):
tenor-111872865

PR #2819, opened 2020-12-30
Make sure Chef Infra Client 15 is listed as deprecated

tas50:
There should always be a release in this spot on the page.


kagarmoe (changes requested):
Let's make the deprecation clear by removing 15.x from the supported table (lines 97-102)

PR #2830, opened 2021-01-01
Collapse ChefDK version info to a single item

tas50:
This matches all the others


kagarmoe:
I get what you're trying to do, but I'd like these entries to stay unique for the time being. I'm open to reconsidering in a few months.

PR #2835, opened 2021-01-05
Remove DK in the support page + update server platforms

tas50:
Making sure the list of products/platforms is correct


kagarmoe (approved):
Thanks!

PR #2857, opened 2021-01-09
Rename Recipe DSL to Chef Infra Language

tas50:
That's what this has always been and then we don't have to overload folks with CS theory and explain what a DSL is.

Signed-off-by: Tim Smith <tsmith@chef.io>

PR #2863, opened 2021-01-12
Break the Infra Language page up into smaller pages

tas50:
This builds on https://github.com/chef/chef-web-docs/pull/2857 which will need to be merged first.

Right now we have a pile of methods all on a single page. This breaks them into smaller pages grouped by function. This is necessary so we can add the dozens of missing methods without making this a 20 page long document.

It's more than just the same content rearranged. There is a significant expansion of some of the content with new methods and tables giving you the parameter values you'd need to be successful.


IanMadd:
We should rename `content/infra_language/overview.md` to `content/infra_language/_index.md`, so there's a page at https://docs.chef.io/infra_language/. Otherwise it just shows a blank page.

tas50:
> We should rename content/infra_language/overview.md to content/infra_language/_index.md

Done

kagarmoe (approved):

PR #2888, opened 2021-01-27
Make the theme simpler

IanMadd:
### Description

This redoes a bunch of Hugo configuration in the background to make our Hugo theme a bit simpler. Otherwise, everything should look and function the same.

- eliminates a redundant `single` layout
- simplifies the infra resources layouts so they use the `baseof` layout instead of using their own having their own head and footer content
- simplifies the 404 layout to do the same
- use partialCached for partials that are always the same on every page 
- removes taxonomies that we don't use, which is all of them
- uses module mounts to mount NPM packages in Hugo instead of moving them around with `cp`
   - updates the Makefile because of this ☝️ 

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


kagarmoe:
Could you explain `use partialCached for partials that are always the same on every page`?

kagarmoe (approved):
tenor-85940493

PR #2929, opened 2021-02-09
Tell swiftype what we want indexed

IanMadd:

### Description

Tell Swiftype the section of each page that we want indexed. This excludes things like the left nav and the TOC.

https://swiftype.com/documentation/site-search/crawler-configuration/content-inclusion-exclusion

This also removes the main docs site landing page and the site map from the index.

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -1,0 +1,6 @@
++++
+title = "Chef Documentation"
+draft = false
+st_robots = "noindex, follow"
+toc = false
++++
--- a/content/attribute_precedence.md
+++ b/content/attribute_precedence.md
@@ -13,7 +13,7 @@ draft = false
 
 {{% node_attribute_precedence %}}
 
-### Examples
+## Examples
 
 The following examples are listed from low to high precedence.
 
--- a/content/environments.md
+++ b/content/environments.md
@@ -80,7 +80,7 @@ formats:
     available when running Chef Infra Client in local mode
 -   As JSON (i.e. a file that ends with `.json`)
 
-### Chef Langage
+### Chef Language
 
 Each environment is defined as a Ruby file (i.e. a file that ends with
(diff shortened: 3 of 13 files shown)


kagarmoe:
Should we add a 'false' on the side nav, or does it only index 'true' and exclude everything else?

IanMadd:
According to the docs, any tag outside of a tag with `data-swiftype-index='true'` will be ignored unless it's specifically added using that same tag. 

https://swiftype.com/documentation/site-search/crawler-configuration/content-inclusion-exclusion#content_inclusion

The alternative is to `data-swiftype-index='false'` to exclude specific sections.
https://swiftype.com/documentation/site-search/crawler-configuration/content-inclusion-exclusion#content_exclusion

kagarmoe:
This doesn't entirely make sense to me; I would think that the default behavior would be either to index or not to index. I'm assuming that the default is to index everything, so let's try excluding the parts we don't want indexed (header, navs, footer) and see if it works as expected.

IanMadd:
I don't understand what's not making sense. The default is to index everything unless you tell it what you want index, then it just indexes those parts. Per the example, the only content that gets indexed is what is inside the div with `data-swiftype-index='true'` :

```

<body>

  This is content that will not be indexed by the Swiftype crawler.

  <div data-swiftype-index='true'>
    <p>
      All content under the above div tag will be indexed.
    </p>
    <p>
      Content in this paragraph tag will be included from the search index!
    </p>
  </div>

  This content will not be indexed, since it isn't surrounded by an include tag.

</body> 
```

kagarmoe (approved):
I am genuinely excited to see how this works!

tenor-232050667

PR #2985, opened 2021-03-03
Image styling

IanMadd:

### Description

A few things: 

- Centers images and adds a box shadow around them.
- Styles the box shadow to handle Windows High Contrast Mode. https://uxdesign.cc/accessible-does-not-mean-ugly-3cb9f65b12f6
- Add a bit more padding and margin around images.
- Changes some images from using the `` tag in markdown to using the `figure` shortcode.
- Adds CSS styling for the `figure` tag.
- Add CSS styling for a `no-shadow` class so we can remove the drop shadow on images that look weird with it.
- Nests some images in HTML tables in a `<p>` tag if they weren't nested inside one before.
- Increase size of images in HTML tables because the added box shadow, margin, and padding made them too small.
- Cleans out some scss junk that isn't used and doesn't need to be there.     
- Updates the `{{% chef_automate_mark %}}` shortcode so it's an HTML shortcode so that the image has the `no-shadow` class.

### Relevant Pages

https://docs.chef.io/automate/iam_v2_guide/
https://docs.chef.io/server/
https://docs.chef.io/automate/reports/

https://deploy-preview-2985--chef-web-docs.netlify.app/automate/iam_v2_guide/
https://deploy-preview-2985--chef-web-docs.netlify.app/server/
https://deploy-preview-2985--chef-web-docs.netlify.app/automate/reports/
https://deploy-preview-2985--chef-web-docs.netlify.app/chef_overview/
https://deploy-preview-2985--chef-web-docs.netlify.app/config_rb_manage/
https://deploy-preview-2985--chef-web-docs.netlify.app/platform_overview/
https://deploy-preview-2985--chef-web-docs.netlify.app/supermarket/

Some of this does look a little weird:
https://deploy-preview-2985--chef-web-docs.netlify.app/server_manage_environments/
https://deploy-preview-2985--chef-web-docs.netlify.app/supermarket_share_cookbook/

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


kagarmoe:
It does look strange when we have shadow-boxes inside of shadow-boxes. Also, the new shadow is much larger than the old. Try reducing the factor that raises the shadow offset and perhaps also the blur radius.

The blur doesn't look great in high-contrast mode. We might want to add styling that turns it into a solid line if high contrast is present.

kagarmoe:
We're going to need to review the docs for pages like this: https://deploy-preview-2985--chef-web-docs.netlify.app/chef_overview/

IanMadd:
So I don't lose this: `(?<!<p>)<img`

IanMadd:
This adds a "no-shadow" class which removes the box shadow can be added to images that are in an html table or to the figure shortcode.

There were several images added to tables without a paragraph tag, which changed how they were styled. So I added a paragraph tag to all those images. 

Images that require a width can be added using the **figure** shortcode which has a width property. I changed a few images so they're added using that shortcode with the width property.

IanMadd:
This is a bit better now, but there are a lot of images that need some custom styling that it will take some time to format all of them properly. For example, of the Chef Manage images should have the "no-shadow" class, but not all of them and the product is deprecated so it might not be worth the effort.

Also the images inside tables. I've edited them so they're all inside <p> tags and they now all have shadows, but maybe they shouldn't. I could edit the CSS so that images in a table don't have shadow, but there may be images that should have a shadow. If that's the case then we should go image by image and remove the shadow from each image.

In any case, I think this is a lot better.

kagarmoe:
Can we remove/reduce the margin around the `images/a2_docs_banner.svg`?

IanMadd:
> Can we remove/reduce the margin around the `images/a2_docs_banner.svg`?

~~That's the image width. I could set it to `width=1000`.~~
Actually that's padding between the img and figure tags

kagarmoe (approved):
Looks great!

PR #3004, opened 2021-03-15
Improve site layout

IanMadd:
### Description

Fixes a few things:

- removes that pesky issue where the page content would scroll left when the right TOC disappeared
- removes `max-width` and adds `fluid` so the page looks better on larger screens
- removes left-nav width because it's not really doing anything
- touch up the javascript resources concatenation in Hugo a little

### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/themes/docs-new/assets/sass/_homepage.scss
+++ b/themes/docs-new/assets/sass/_homepage.scss
@@ -3,7 +3,6 @@
 
   &-col {
     padding-right: rem-calc(12px);
-    max-width: 1000px;
     min-height: 100vh;
     margin-bottom: 30px;
     &:focus {
--- a/themes/docs-new/assets/sass/_variables.scss
+++ b/themes/docs-new/assets/sass/_variables.scss
@@ -94,3 +94,7 @@ $header-styles: (
     'h6': ('font-size': 16),
   )
 );
+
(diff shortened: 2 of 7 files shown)


kagarmoe (approved):

IanMadd:
We've got some issues with page menu titles in the left nav. Some are nested several layers deep and have long titles (mostly resources) and currently the ends of their titles hidden by the main page content.

Screen Shot 2021-03-17 at 4 40 47 PM

This PR currently allows the left nav to scroll sideways so the user can see the full page title, but that seems a bit sloppy. 

We could make the nav wider, but we'd have to make it about 300px wide (currently it's 250px).

We could let the page titles break to a new line, but the resource names break in a weird places. 
Screen Shot 2021-03-17 at 5 04 17 PM

Or maybe change the font size and letter spacing for nested elements, which would look something like this:
Screen Shot 2021-03-17 at 5 24 57 PM

There are also some suggestions here: https://ux.stackexchange.com/questions/60830/handling-long-text-wrapping-in-multi-level-navigation-sidebars

https://css-tricks.com/handling-long-unexpected-content-css/

PR #3046, opened 2021-04-08
Chef Habitat Release Notes

kagarmoe:
### Description

Adds the page for pulling in the RN. The historical RN are already in the S3 bucket. 

* Need to add versions to `onmitruck.chef.io/stable/habitat/versions/all`
### Issues Resolved

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


IanMadd:
List of Hab versions: https://bldr.habitat.sh/v1/depot/pkgs/core/hab/versions

kagarmoe:
@IanMadd Should I go ahead and close this?

IanMadd:
We're somewhat dead in the water on this. The hab versions list that I linked to above doesn't accept CORS requests and no one seems to know if it's possible to change that. 🤷 There's also no `version` metadata with the expeditor workload that updates the hab Hugo module, so we can't create a JSON/YAML file that's curated by Expeditor. At this point the only thing I can think of is manually maintaining a JSON/YAML file. But then our jobs are to automate the world so why would we stoop so low as to manually maintain something that should be done by a :robot:?

PR #3084, opened 2021-04-27
Add 17.0 and 15.17 release notes

tas50:


tas50:
We can rereview this later, but this needs to be live for the release that is out.

PR #3112, opened 2021-05-03
CanonifyURLs

kagarmoe:

### Description

Add canonical URLs to help SEO.

### Issues Resolved

#3112 

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


config.toml
```diff
@@ -1,4 +1,4 @@
-baseURL = "/"
+baseURL = "https://docs.chef.io/"
```
tas50, line 1: Does this break the local build?

kagarmoe: I expected it to, but it did not. However, it also didn't create a canonical URL.

IanMadd: Not that this is in the Hugo docs anywhere, but the Hugo server knows if it's a local build and will use localhost instead of baseURL.

kagarmoe:
@IanMadd Will this break the world in production?

IanMadd:
It's exhibiting a little weirdness. The external link icons are missing. Not sure why. I'll have to mess with that. I think those are just called from Fontawesome.

What's weird about this is that it's not finding the fonts, either fontawesome or the fonts specified in sass, but all the other styling is working. And even worse, these things work locally but not when deployed, which is the opposite of what you might think.

Another problem this is creating is that it's setting the URL for the sass style sheets as an absolute URL to docs.chef.io, so if we make style changes, we won't be able to preview them.

IanMadd (approved):
tenor-255891935

PR #3113, opened 2021-05-03
Add descriptions to improve SEO

kagarmoe:

### Description

We should add "description" fields to the frontmatter to improve SEO. Hugo has a neat `.Summary` page variable that grabs the first 70 words of the document. This PR tries to get it to work.

### Definition of Done

- [ ] Adds description field and `.Summary` to archetypes.
- [ ] Tries it out on `api_omnitruck`.

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


IanMadd:
Changes I made:
- created a new partial for summaries because it's getting a bit long
- change summary length to 120 characters cause that's apparently acceptable
- added ability to read data files for infra resource and cookstyle cop summaries
- added the ability to use a  page `description` parameter in the description meta tag.

Things that still need to be done:
- api pages have no meta description content, but they can get it if we add a `description` to the frontmatter of each api page
- desktop resource pages also have no meta description content. This one's more complicated.

IanMadd (approved):

kagarmoe:
`summaryLength = 120` means 120 words. I changed it to 20 words. It doesn't look like we have the option for characters (which would be better) https://github.com/gohugoio/hugo/blob/b660ea8d545d6ba5479dd28a670044d57e5d196f/helpers/content.go#L259

kagarmoe:
The rest of the work looks great! Good thinking on the shortcode.

Issue #3063, opened 2021-04-15
[BUG] Chef Backend Deprecation Note


kagarmoe:
## General details

Add a deprecation note for Chef Backend

Link to page and section that needs updated:
https://docs.chef.io/server/upgrade_server_ha_v2/

Reported by  @sirajrauff
https://chefio.slack.com/archives/C07JP2HN1/p1618507246041300

## Product

- [ ] Chef Automate
- [ ] Chef Habitat
- [ ] Chef Infra Client
- [x] Chef Infra Server
- [ ] Chef InSpec
- [ ] Other________________

## Describe the update that needs to be made. Be as specific as possible:

Analysis: What's schedule for Backend EOL
Definition of Done: Add a deprecation note about Backend

PR #3120, opened 2021-05-06
Deprecation Note for Chef Backend 

dishanktiwari2501:
Deprecation note added for Chef Backend

**Reference PR:** https://github.com/chef/chef-web-docs/issues/3063

**Link to page and section that will be updated:**
https://docs.chef.io/server/upgrade_server_ha_v2/

Reported by @sirajrauff
https://chefio.slack.com/archives/C07JP2HN1/p1618507246041300

### DIFF
--- a/themes/docs-new/layouts/shortcodes/EOL_backend.md
+++ b/themes/docs-new/layouts/shortcodes/EOL_backend.md
@@ -1,0 +1,7 @@
+{{% EOL_manage %}}
+
+Chef Backend is [deprecated](/versions/#deprecated-products-and-versions) and no longer under active development. Contact your Chef account representative for information about upgrading your system.
+
+This document is no longer maintained.
+
+{{% /EOL_manage %}}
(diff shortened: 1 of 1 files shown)


IanMadd:
Hugo doesn't let you call a shortcode from another shortcode, it just doesn't work like that. It's possible to nest shortcodes, but not by name.

So what you probably want to do is remove `{{% EOL_manage %}}` from lines 1 and 7, and then after this is merged, invoke this from `server/upgrade_server_ha_v2.md` by adding `{{% EOL_backend %}}` at the top of that page.

IanMadd:
What you might do is something like this:

Remove the {{% EOL_manage %}} from the 1st and 7th lines.

Then after this is merged, add this to the automate page with:

```
{{< note >}}
{{% EOL_backend %}}
{{< /note >}} 
```

or 


```
{{< warning >}}
{{% EOL_backend %}}
{{< /warning >}} 
```

IanMadd (approved):
tenor-258564458

Issue #3124, opened 2021-05-06
[INTAKE] Custom Resources Documentation needs a rethink


damacus:
# Documentation project plan

- Custom Resources currently sprawl across multiple pages
- We should rewrite them from the ground up taking what is required from the old documentation
- Separate out upgrading from definitions/LWRPS into their owns docs
- Write a simple getting started guide
- Write a more complex guide with gotchyas (e.g. accumulators)
- A complete DSL guide with a full terms of reference

## Product

Chef Infra Client

Brief product description: [3-4 lines]

Product manager: @tas50

### Target users

- [x] Sysadmins
- [x] Developers
- [x] Internal engineers
- [ ] Business decision-makers

### Documentation size estimate

~ 5 pages

### Reviewers for content

- @tas50
- @lamont-granquist
- @Xorima / @bmhughes

## Actions

- [ ] Document migration from LWRP to Custom Resources
- [ ] Document migration from definitions to Custom Resources
- [ ] new\_resource vs current resource
- [ ] Libraries
- [ ] action\_class
- [ ] Accumulators

## Review Exiting Community Documentation

- [ ] [Chef Custom Resources - Missing Documentation](https://blog.backslasher.net/chef-custom-resources.html)

PR #3147, opened 2021-05-14
[GH-3124] Remove definition documentation

damacus:
Remove mention of definitions

We have not recommended the user of definitions for a long time. Custom resources have been out for 4+ years, and we should deprecate/remove this part of the documentation

## Related tickets

- Relates to #3124


tas50:
We probably want to keep a document on migrating to definitions. I'm not sure how that would fit in our existing docs structure, but it's an important document to offer to folks trying to get off old patterns. The glossary entry should also stay until we kill off support entirely.

tas50:
Do we still need this since the new custom resource PR includes the definition rework?

damacus:
They're related but separate. #3185 Adds the migration docs that we need to get rid of this. But it can be merged before/after, if we want.

tas50 (approved):

kagarmoe (approved):

PR #3149, opened 2021-05-14
Kill off recipe DSL in a few more places

tas50:
This is a term that should never be used again


kagarmoe (approved):
tenor-93213506

PR #3168, opened 2021-05-20
Remove diff property from remote_file

tas50:
This is used internally only


kagarmoe (approved):
Thanks!

IanMadd (approved):

PR #3170, opened 2021-05-20
Remove stove from the Supermarket sharing page

tas50:
We no longer ship stove in Workstation and don't recommend using it.


kagarmoe (approved):
Thanks!

PR #3173, opened 2021-05-20
Update the requirements for Supermarket + airgap

tas50:
- List the actual distros we support
- Remove the mention of Chef Server 12+. That bar is so low it doesn't
  need to be mentioned
- Remove some default knife config from the airgap page


kagarmoe (approved):
Thanks!

PR #3239, opened 2021-06-16
Wrapper Cookbooks

kagarmoe:

## Description

Wrapper cookbooks are really important in Chef.
Let's explain the wrapper cookbook pattern outside of the context of installing a private Supermarket.

The first draft is a copy-paste of the Supermarket section. Let's make a generic example and iterate from there.

This also cleans up some of the Markdown on the original Supermarket page (WIP)

## Definition of Done

I'd like to merge this by July 1.

## Issues Resolved

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


content/cookbook_wrapper.md
```diff
@@ -0,0 +1,149 @@
...
+
+{{< /note >}}
+
+### Upload the Wrapper
+
+The wrapper cookbook around the `SOMETHING` cookbook must be uploaded to the Chef Infra Server, along with any cookbooks against which the `SOMETHING` cookbook has dependencies.
+
+To upload the cookbooks necessary to install Chef Supermarket, do the following:
+
+1. Install Berkshelf:
```
sam1el: I think we may want to start removing references to berkshelf. It's not really required, it's been mostly deprecated for years in lieu of policyfile and, those cookbooks can still be uploaded with their dependencies to supermarket

content/install_supermarket.md
```diff
@@ -288,137 +207,113 @@ node.override['supermarket_omnibus']['config']['fqdn'] = your_node_public_ip
...
-against which the `supermarket-omnibus-cookbook` cookbook has
-dependencies.
+The wrapper cookbook around the `supermarket-omnibus-cookbook` cookbook must be uploaded to the Chef Infra Server, along with any cookbooks against which the `supermarket-omnibus-cookbook` cookbook has dependencies.
 
-To upload the cookbooks necessary to install Chef Supermarket, do the
-following:
+To upload the cookbooks necessary to install Chef Supermarket, do the following:
 
-1.  Install Berkshelf:
+1. Install Berkshelf:
```
sam1el, line 214: same as above

content/cookbook_wrapper.md
kagarmoe: done!

kagarmoe: done

content/cookbook_wrapper.md
```diff
@@ -0,0 +1,145 @@
...
+    ```
+
+3. Save and close the `/recipes/default.rb` file.
+
+{{< note >}}
+
+If you are running your private Supermarket in AWS, you may need to set an additional attribute for the node's public IP address:
+
+```ruby
+node.override['supermarket_omnibus']['config']['fqdn'] = your_node_public_ip
```
tas50, line 114: Since we've already said we're talking about AWS we could use the cloud attribute here.

content/cookbook_wrapper.md
```diff
@@ -0,0 +1,145 @@
...
+
+    ```ruby
+    include_recipe 'SOMETHING'
+    ```
+
+    This ensures that the `default.rb` file in the `SOMETHING` is run.
+
+## Define Attributes
+
+Define the attributes for the Chef Supermarket installation and how it connects to the Chef Infra Server. One approach is to hard-code attributes in the wrapper cookbook's `default.rb` recipe. A better approach is to place these attributes in a data bag, and then reference them from the recipe. For example, the data bag could be named `apps` and then a data bag item within the data bag could be named `supermarket`.
```
damacus, line 67:
```suggestion
Define the attributes for the Chef Supermarket installation and how it connects to the Chef Infra Server. One approach is to hard-code attributes in the wrapper cookbook's `default.rb` recipe.
```

Data bags are *fine*, but I wouldn't say they're the best way of doing this as they're not version controlled.

damacus:
~Should we be mentioning the [omnibus-supermarket cookbook](https://supermarket.chef.io/cookbooks/supermarket-omnibus-cookbook) instead of the [deprecated supermarket cookbook](https://supermarket.chef.io/cookbooks/supermarket)?~

~That said, if we're wrapping a modern cookbook, we should be wrapping a cookbook with custom resources in it. A good example of a modern wrapper would be something like the [Apache2 test cookbook](https://github.com/sous-chefs/apache2/tree/main/test/cookbooks/test).~

~Mentioning databags and extra attributes at this point (which I think is  a beginner friendly page?) might just be making this page a lot to take in.~

On second thoughts I think we should just pull this PR in and iterate over the content. The layout is right but a lot of the content is just plain wrong.

damacus (approved):

kagarmoe:
Revisit when we have cycles to update.

PR #3252, opened 2021-06-25
Style changes for responsive images

kagarmoe:

## Description

The use of `box-shadow` in our styling meant that we had to size figures specifically, (width="500") and we couldn't use percentages without the entire area seeming to jump off the page. Removing the shadows & adding a lightgrey around the image in a figure fixes this.

This PR also removes the padding and margin that was constraining the images to even smaller sizes.

The problem: 

`{{< figure src="/images/automate/add-chef-organization-popup-menu.png" width="500" alt="Add Chef Organization Form" width="60%">}}`

Screen Shot 2021-06-25 at 4 46 45 PM

This solution:

`{{< figure src="/images/automate/add-chef-organization-popup-menu.png" alt="Add Chef Organization Form" width="60%">}}`

Screen Shot 2021-06-25 at 5 01 29 PM

## Issues Resolved

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/_vendor/github.com/chef/automate/components/docs-chef-io/content/automate/chef_infra_server.md
+++ b/_vendor/github.com/chef/automate/components/docs-chef-io/content/automate/chef_infra_server.md
@@ -39,11 +39,11 @@ The _Chef Infra Server_ panel starts out with an empty list of servers.
 
 To add existing Chef Infra Servers to the Chef Automate infrastructure, select **Add Chef Server** which will request the name, FQDN, and IP address of your Chef Infra Server:
 
-{{< figure src="/images/automate/add-chef-server-popup-menu.png" width="500" alt="Add Chef Server Form">}}
+{{< figure src="/images/automate/add-chef-server-popup-menu.png" width="60%" alt="Add Chef Server Form">}}
 
 Chef Automate will warn you if you enter an invalid FQDN or IP address:
 
-{{< figure src="/images/automate/add-chef-server-popup-menu-with-error.png" width="500" alt="Add Chef Server Form">}}
+{{< figure src="/images/automate/add-chef-server-popup-menu-with-error.png" width="60%" alt="Add Chef Server Form">}}
 
 Once you are finished, select **Add Chef Server** and you will see your server in the list of Chef Infra Servers.
 
--- a/themes/docs-new/assets/sass/partials/_grid.scss
+++ b/themes/docs-new/assets/sass/partials/_grid.scss
@@ -27,12 +27,9 @@
     p img {
       display: block;
       margin: rem-calc(32) auto;
-      -webkit-box-shadow: $box-shadow;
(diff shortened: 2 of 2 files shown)


IanMadd:
This turned into a crazy deep dive on my part.

First this highlights a problem with our use of the figure shortcode. Some of them have a width defined in pixels, but without specifying that they're pixels ==> `width="500"` should be `width="500px"`. Hugo ignores this b/c it doesn't know if it's pixels or percent, so the image width is 100%. Read on.

Second, ignoring your PR for a second. Setting `width="60%"` sets the width of the image to 60% of the containing figure tag, which is already has a max width of 75%. That's why that looks weird, it's 60% of 75%, and the border is at 75%.

This PR creates a few problems. 

- Hugo wraps all content from the figure shortcode inside the figure tag. So the caption and title, if they exist, will be outside the border, and they look weird that way.

- Also many images could use padding because they don't have any in the image, so the border is right up against the edge of the image. 

These look a bit weird: 
https://deploy-preview-3252--chef-web-docs.netlify.app/chef_overview/#chef-infra-components
https://deploy-preview-3252--chef-web-docs.netlify.app/style_markdown/#figure-shortcode

Maybe the underlying issue is that we've been inconsistent about creating images and standardizing things like how big an image is, that the text needs to be readable but not too large, and should the image have a white margin around its edge as part of the image. A lot of our images drive me bonkers because of how inconsistent they are, some are gigantic with huge text, and some are tiny and illegible, some have white padding in the image, many don't.

Anyway, the figure shortcode accepts a class value, so we can create a class if we need images to be set to a particular width, or some other styling.

Also, the figure shortcode is weird. If you pass in a width, it applies it to the enclosed img tag, if you pass in a class, it applies it to the figure tag. Weird.

kagarmoe:
I completely agree that we need to standardize and to enforce those standards.

I set images that are inside of figures to 100%, which means that they should be the size we set on the figures. 

The images that you point out are claustrophobic--Do you think we need borders at all?

The problem that I'm trying to solve in this PR is that using responsive sizes in images cause a large chunk of the screen to jump out at the user--the entire figure space, rather than just the images. I think this PR gets us most of the way. I agree that we need classes.

Ignoring this PR for another moment, it is perhaps time to fully commit to Hugo's image processing:
https://gohugo.io/content-management/image-processing/#the-image-page-resource

Related blog post: https://alexlakatos.com/web/2020/07/17/hugo-image-processing/

IanMadd:
What do you mean when you say "responsive sizes"? The images as they currently exist are responsive, in that they have a max-width so they shrink as the page width gets smaller. Do you want to this max width to be adjustable so a narrower image has a smaller max-width, or a smaller width entirely?

kagarmoe:
Working towards the dream of responsive on different devices

PR #3330, opened 2021-08-11
Infra Nav Update

kagarmoe:
## Description

The number and variety of things in the "Cookbook Reference" section is bewildering. This PR reorganizes the nav entries so that the headers and entries are both clear and necessary.

## Related PRs
Uses content from https://github.com/chef/chef-web-docs/pull/3329

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

PR #3331, opened 2021-08-12
Chef Infra Client Nav Reorg

kagarmoe:
## Description

Reorganize the Infra nav

## Definition of Done

- [x]  Consistent headers
- [ ] Entries in reasonable order
- [ ] A/B test

## Related PRs

Based on https://github.com/chef/chef-web-docs/pull/3330

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


tas50:
A few things:

- Install is in the menu twice
- The knife stuff from the extensions section should go in workstation since that's not an infra feature anymore
- The way we're going with the product I wouldn't call InSpec language an extension at this point. It's just something we should be linking out from compliance phase docs to say "and this is how you will write compliance code"
- Policyfile.rb is a reference doc but it's in the policy section. We also have some weird overlap here with the Policyfile.rb content in the Policyfiles page
-

kagarmoe:
I've addressed all of Tim's concerns, except for Policyfiles, which we'll need to work through at some time in the future.

IanMadd (approved):

PR #3338, opened 2021-08-17
Remove incorrect note about supermarket.rb config file

tas50:
It gets created by default now


kagarmoe (approved):
tenor-117597348

PR #3419, opened 2021-09-14
Fix a few more release notes mistakes

tas50:
Formatting and put a header in the right place


IanMadd (approved):

kagarmoe (approved):
tenor-264945957

I have no idea what this _very specific gif_ that I found means, but I'm using it because it **exists**.

PR #3486, opened 2021-10-13
Add symbol verifiers to `verify` property of `File` resource

gogsbread:
Signed-off-by: Antony Deepak Thomas <antonydeepak@gmail.com>

## Description
Add `:json`, `:yaml` and `:systemd_unit` symbol verifier documentation to `File` resource.

## Definition of Done
Tested in local development and Hugo template renders correctly

## Check List

- [ ] Spell Check
- [x] Local build
- [x] Examine the local build
- [ ] All tests pass

### DIFF
--- a/data/infra/resources/file.yaml
+++ b/data/infra/resources/file.yaml
@@ -29,7 +29,7 @@ syntax_full_code_block: "file 'name' do\n  atomic_update              true, fals
   \ false\n  manage_symlink_source      true, false\n  mode                      \
   \ String, Integer\n  owner                      String, Integer\n  path        \
   \               String # defaults to 'name' if not specified\n  rights         \
-  \            Hash\n  verify                     String, Block\n  action        \
+  \            Hash\n  verify                     String, Block, Symbol\n  action        \
   \             Symbol # defaults to :create if not specified\nend"
 syntax_full_properties_list:
 - '`file` is the resource.'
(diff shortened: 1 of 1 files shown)


tas50 (approved):

gogsbread:
@tas50 lmk if you want to fix any of the existing CI errors. The spellchecker & linter errors don't seem to be related to this patch. The Netlify seems to take forever as well.

kagarmoe:
Ignore the Markdown and spellcheck errors for now. I'm in the process of implementing those tests. Should be fixed this week.

data/infra/resources/file.yaml
```diff
@@ -260,46 +260,98 @@ properties_list:
...
-      as a block, it returns `true` or `false`. When specified as a
-
-      string, it is executed as a system command. It evaluates to `true`
-
-      if the command returns 0 as its exit status code and `false` if the
-
-      command returns a non-zero exit status code.'
+      The types for this property are a block, string, or a symbol. 
+      
+      * A block returns `true` or `false`.
```
kagarmoe: @tas50 Can I do bulleted lists in resource YAML files?

tas50: it's just a big string in the resource so you can put whatever you want in it.

gogsbread:
hmm.. looks like remote build is failing; it passed locally though. I'll see what is happening

data/infra/resources/file.yaml
```diff
@@ -260,46 +260,98 @@ properties_list:
...
+      
+        ```ruby
+        file 'foo.yaml' do
+            content "--- foo: 'foo-"
+            verify :yaml
+        end
+        ```
+        
+        If a string, block or symbol returns `false`, then the Chef Infra Client run
+        stops and raises an error is raised.
```
gogsbread: correct verbiage

gogsbread:
> hmm.. looks like remote build is failing; it passed locally though. I'll see what is happening

Corrected the syntax a bit to see if `hugo` is happy. Not sure why the build command is failing; maybe it is using a different compiler than the one I use locally??

kagarmoe:
@gogsbread I don't think the problem is on your end. I'm going to admin merge this. I *think* that the workflow won't assign your branch VMs for testing because you're working from a fork of the repo.

kagarmoe (approved):

PR #3505, opened 2021-10-22
Rework the description of a policyfile

tas50:
This was just wrong. Flat out.


themes/docs-new/layouts/shortcodes/policyfile_summary.md
```diff
@@ -1,5 +1 @@
-A Policyfile is an optional way to manage role, environment, and
-community cookbook data with a single document that is uploaded to the
-Chef Infra Server. The file is associated with a group of nodes,
-cookbooks, and settings. When these nodes perform a Chef Infra Client
-run, they utilize recipes specified in the Policyfile run-list.
\ No newline at end of file
+A Policyfile is a way to create immutable collections of cookbooks, cookbook dependencies, and attributes defined in a single document that is uploaded to the Chef Infra Server. The Policyfile is then associated with a group of nodes. When these nodes perform a Chef Infra Client run, they utilize recipes specified in the Policyfile.
```
kagarmoe, line 1: I don't understand how `The Policyfile is then associated with a group of nodes.` happens

kagarmoe:
```suggestion
A Policyfile is a single document for defining immutable collections of cookbooks, cookbook dependencies, recipes, and attributes. You write your Policyfile on your Chef Workstation and upload it to the Chef Infra Server. [The Policyfile is then associated with a group of nodes.] During a Chef Infra Client run, these nodes use the recipes defined in the Policyfile.
```

tas50: Policyfile is a bit loaded as a term and we should shy away from talking about the specific file / document. The policyfile doc generates a policyfile lock and that policy object is applied to a node or a group of nodes via policy group.

kagarmoe: I think we should explain it completely in at least once. The place could be here, or it could be somewhere else.

kagarmoe (approved):

PR #3522, opened 2021-10-31
Fix mistake in "force" warning syntax

### LINKED ISSUE #3521: [BUG] User resource attribute force warning not rendering
## General details

Something's wrong with the rendering of this warning. The [warning is represented in this doc file](https://github.com/chef/chef-web-docs/blob/1b58656d72b0b101b4ed2a07bfe352796f498046/data/infra/resources/user.yaml#L89-L96) and should render in the [`force` section of this rendered doc page](https://docs.chef.io/resources/user/#properties) but it doesn't.

chef doc page rendering incorrectly

## Product

[ ] Chef Automate
[ ] Chef Habitat
[x] Chef Infra Client
[ ] Chef Infra Server
[ ] Chef InSpec
[ ] Other________________

## Describe the update that needs to be made. Be as specific as possible:

[In other pages warnings are rendered in HTML instead of markdown and it works](https://github.com/chef/chef-web-docs/blob/1b58656d72b0b101b4ed2a07bfe352796f498046/data/infra/resources/execute.yaml#L459-L465)

[Here's an example of an `info` box in markdown that does work](https://github.com/chef/chef-web-docs/blob/1b58656d72b0b101b4ed2a07bfe352796f498046/data/infra/resources/git.yaml#L110-L117)

Based on the git page I've opened #3522 which may fix this

gene1wood:
This should be a hash with a key of `markdown` not an array containing an element that's a hash with a key of `markdown`
Fixes #3521

## Description

This should fix the bug reported in #3521 based on how the working `note` field in the [git docs](https://github.com/chef/chef-web-docs/blob/1b58656d72b0b101b4ed2a07bfe352796f498046/data/infra/resources/git.yaml#L110-L117) is done

## Issues Resolved

* #3521 

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/data/infra/resources/user.yaml
+++ b/data/infra/resources/user.yaml
@@ -87,7 +87,7 @@ properties_list:
 
       action.'
   - warning:
-    - markdown: 'Using this property may leave the system in an inconsistent state.
+      markdown: 'Using this property may leave the system in an inconsistent state.
 
         For example, a user account will be removed even if the user is
 


IanMadd (approved):

IanMadd:
Thanks for the fix @gene1wood!

gene1wood:
@IanMadd Do I need to amend my commit with the DCO signed-off-by or with the "Obvious fix" note?

IanMadd:
It's such a minor change that I don't think it's worth amending.

gene1wood:
@IanMadd ~Actually maybe ignore my question as it looks like this change doesn't result in the fix I was hoping. Looking at the test deploy the problem appears to still be present : https://docs.chef.io/resources/user/#properties~ 

Nevermind, though the fix didn't seem to display in the test site, on the live site the fix worked. So all good. Thanks for the review and merge!

PR #3524, opened 2021-11-02
Fix malformed warnings in two resources

gene1wood:
This is the same bug as in #3521 that I fixed in #3522 but on two other doc pages.
These doc pages render an empty Warning block because the YAML contains an array of hashes
instead of a hash. This should be a hash with a key of `markdown` not an array containing an element that's a hash with a key of `markdown`.

## Description

Fixes empty Warning blocks

## Issues Resolved

## Related PRs

* #3522

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/data/infra/resources/dsc_script.yaml
+++ b/data/infra/resources/dsc_script.yaml
@@ -160,7 +160,7 @@ properties_list:
   required: false
   description_list:
   - warning:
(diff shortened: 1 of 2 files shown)


IanMadd (approved):
Thanks!

PR #3555, opened 2021-11-15
Removed the outdated note

### LINKED ISSUE #3371: Remove outdated note about running Infra Server with 4GB
## General details

Link to page and section that needs updated:

https://docs.chef.io/chef_system_requirements/

## Product

[ ] Chef Automate
[ ] Chef Habitat
[ ] Chef Infra Client
[x] Chef Infra Server
[ ] Chef InSpec
[ ] Other________________

## Describe the update that needs to be made. Be as specific as possible:

We need to remove the note that tells people they can run on 4GB of ram. This is not a thing we should suggest at this point even with small server installs.

dishanktiwari2501:
## Description

We need to remove the note that tells people they can run on 4GB of ram. This is not a thing we should suggest at this point even with small server installs.

## Definition of Done

The note has been removed.

## Related PRs

- https://github.com/chef/chef-web-docs/issues/3371

closes #3371


IanMadd (commented):
This should remove lines through the second closing div tag on line 31/25.

IanMadd:
This should remove the text of the note too. "The RAM requirement..."

kagarmoe (approved):

PR #3325, opened 2021-08-11
Make feedback buttons visible

kagarmoe:
Signed-off-by: kagarmoe <kgarmoe@chef.io>

## Description

Let's test if moving the feedback buttons increases the feedback.

1. I'm not convinced this will report the right information
2. We may want to improve the styling.

## Issues Resolved

[List any existing issues this PR resolves, or any Discourse or
StackOverflow discussion that's relevant]

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

PR #3564, opened 2021-11-17
Updated the new feedback section in the doc

dishanktiwari2501:
Signed-off-by: Dishank Tiwari <dtiwari@progress.com>

## Description

A New feedback section has been added with thumbs up and down buttons.

## Related PRs

- https://github.com/chef/chef-web-docs/pull/3325

## Check List

- [x] Spell Check
- [x] Local build
- [x] Examine the local build
- [x] All tests pass

### DIFF
--- a/themes/docs-new/assets/sass/_mixins.scss
+++ b/themes/docs-new/assets/sass/_mixins.scss
@@ -1,8 +1,4 @@
 @import 'foundation-sites/scss/util/mixins';
-@import "vendor/fontawesome/fontawesome.scss";
-@import "vendor/fontawesome/solid.scss";
-@import "vendor/fontawesome/icons.scss";
-@import "vendor/fontawesome/icons.scss";
 
 // Font smoothing mixin
 @mixin font-smoothing() {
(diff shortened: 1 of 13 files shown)


kagarmoe:
The border doesn't go completely around the widget. Is this intentional?

Screen Shot 2021-11-30 at 3 43 08 PM

themes/docs-new/layouts/_default/infra_resource.html
```diff
@@ -750,6 +750,6 @@ <h2 id = "examples">Examples</h2>
         {{ end }}
       </div>
     </div>
-    {{ partialCached "feedback" . }}
```
IanMadd, line 753: We should keep this.

themes/docs-new/assets/sass/_mixins.scss
```diff
@@ -1,4 +1,8 @@
 @import 'foundation-sites/scss/util/mixins';
+@import "vendor/fontawesome/fontawesome.scss";
+@import "vendor/fontawesome/solid.scss";
+@import "vendor/fontawesome/icons.scss";
+@import "vendor/fontawesome/icons.scss";
```
IanMadd: I don't think this is necessary. We're already bringing in fontawesome in the _typography file, so these resources should be available without this.

IanMadd: Also this section is for mixins, so it's not the right place to include these anyway.

PR #3185, opened 2021-05-26
[3124] Custom Resources rewrite

damacus:
- Adds a new menu point for Custom Resources
- Move all custom resource docs to the custom_resource folder
- Move all the DSL pages under custom_resources/dsl so it's easier to maintain
- Add LWRP to Custom Resources guide
- Add Definition to Custom Resources guide
- Remove two shortcodes that should be full pages (resource_name & ruby_type)
- Add DSL page on: coerce, action_class & accumulators

PR #3586, opened 2021-11-30
Kg/modernization

kagarmoe:
## Description

In the course of revising #3185 it became clear that we need extensive revisions to the notes and migration docs.
New Chef Infra admins inherit old Chef Infra systems and don't have any way of understanding the code that they are reading and its conscequences. This `modernization` doc is a handbook of old Chef and what to do about it.

## Definition of Done

- [ ] Call this "Modernization"
- [ ] Document both HWRP and LWRPs as sections
- [ ]  update `converge_by` must always be wrapped (correctness)
- [ ]  `def whyrun_supported`? is no longer required -> Delete def bloc
- [ ]  Document `use_inline_resources`
- [ ]  Document `updated_by_last_action`
- [ ]  Document

    ```text
    # NEVER use `def action_run` here -- you defeat use_inline_resources and will break notifications if you do
    # If you don't understand how use_inline_resources is built and why you have to use the `action` method, and what the implications are and how resource notifications
    # break if use_inline_resources is not used and/or is broken, then you should really not be using library providers+resources.  You might feel "closer to the metal",
    # but you're now using a chainsaw without any guard...` not in the code block
    ```

- [ ] Don't document helpers here
- [ ]  `def action run` block superficially works but needs to be converted to `action :run do`
- [ ]  Don't use Definitions...change them to example

## Issues Resolved

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


content/custom_resources_notes.md
```diff
@@ -195,6 +209,8 @@ please stop writing actions this way.
 
 THIS IS CORRECT:
```
kagarmoe, line 210: This is no longer correct.

content/lwrp_to_custom_resources.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Convert files to Custom Resources Layout
+
+HWRPs consist of two library files: a resource and a provider for that resource.
+
+```text
+|- libraries
+  |- provider_rvm_ruby.rb
+  |- resource_rvm_ruby.rb
+```
```
kagarmoe, line 32: from @tas50 `If there are two library files it's not an LWRP. It's an HWRP. LWRPs have a file in /resources and a file in /providers.`

PR #3092, opened 2021-04-28
Unified Mode

lamont-granquist:
Signed-off-by: Lamont Granquist <lamont@scriptkiddie.org>

### Description

[Please describe what this change achieves]

### Issues Resolved

Parent branch of https://github.com/chef/chef-web-docs/issues/3111

### Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

PR #3596, opened 2021-12-04
Clarify unified_mode docs

jasonwbarnett:
## Description

I think this update reflects reality a bit clearer. Looking for input if I am misunderstanding the current state of unified_mode.

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


themes/docs-new/layouts/shortcodes/unified_mode.md
```diff
@@ -1,6 +1,4 @@
-Unified mode combines the compile and converge stages of the Chef Infra Client run into one phase. In unified mode, Chef Infra Client compiles and converges a custom resource in one pass and in the order that the code is compiled, from the beginning to the end.
-
-Unified mode is only available on custom resources. Chef Infra Client processes all other resources with separate compile and converge phases.
+Unified mode combines the compile and converge stages of the Chef Infra Client run into one phase. When a resource has unified mode enabled, the Chef Infra Client compiles and converges that resource's action block in one pass and in the order that the code inside that block is compiled, from the beginning to the end.
 
 In Chef Infra Client 17 (April 2021) and some earlier versions, the default behavior is `unified_mode false`. Enable unified mode on a custom resource with `unified_mode true`. Chef Infra Client displays a deprecation message with `unified_mode false`.
```
jasonwbarnett, line 3: > In Chef Infra Client 17 (April 2021) and some earlier versions, the default behavior is `unified_mode false`. Enable unified mode on a custom resource with `unified_mode true`. Chef Infra Client displays a deprecation message with `unified_mode false`.

@tas50 @lamont-granquist is this even right? Isn't unified_mode now enabled on every built-in resource?

tas50: It's in all the built-in stuff but that doesn't impact the end user. That's all under the hood.

jasonwbarnett: I think this doc needs some love to clarify the difference between unified_mode as it is expected to behave in Chef 18 with unified_mode during recipe compile time versus unified_mode within a resource itself.

jasonwbarnett: Maybe I'm biased and looking at this entire doc from a forward looking perspective (what will the state of the world be in Chef 18). 

To give a little more context as to how this all came up... I was speaking with a coworker regarding us enabling `unified_mode` on all of our custom resources. It led us to this document and my coworker's interpretation of this document is that somehow custom resources with `unified_mode` enabled would behave differently from built-in resources. He starting having all sorts of concerns for notifications between resource within a recipe itself (not the resource action block). This led us to some confusion around whether or not we need to be concerned about the use of the custom resource opposed to just strictly concerning ourselves with the body of the action blocks.

I won't say our experience is common, but wanted you to at least have the perspective to understand if we're unique and looking at this in a strange way or something else.

P.S. In general it's been a relatively big struggle for me personally to understand and follow `unified_mode`'s current state across the entire Chef ecosystem. i.e. when will unified_mode impact the chef-client run as a whole versus just the resource action blocks? Does enabling unified_mode actually do anything beyond the scope of the action block within the resource? etc.

It really wasn't until yesterday/today where I forced myself to walk through every mention of `unified_mode` in the code base and dug into several sources of docs (this one and the open PR from @lamont-granquist https://github.com/chef/chef-web-docs/pull/3092) that I started to get an understanding.

kagarmoe:
Hi @jasonwbarnett ~could you give me access to your branch?~ Not needed, thanks!

kagarmoe (approved):
From the conversation, I thought that this change was much more involved. Thanks for your input!

tenor-215403058

jasonwbarnett:
@kagarmoe thanks for engaging and getting this merged in. please let me know if you ever want any feedback on any of the unified_mode docs in the future.

Issue #3621, opened 2021-12-21
Automate Manage release notes


tas50:
**Is your feature request related to a problem? Please describe.**

Currently release notes for Chef Manage are not up. Only old ones in a hand crafted doc.

**Describe the solution you'd like**

We should utilize the same automation we're using in other products to push release notes on deploy

**Describe alternatives you've considered**

Whatever you think is best

IanMadd:
- [x] Add manage release notes pages to packages.chef.io
- [x] Add Manage release notes to chef-web-docs
- [x] Add expeditor config to chef/chef-manage to handle publishing and announcing releases

IanMadd:
Closed with the above PRs ☝️

Issue #3519, opened 2021-10-28
Feature Request: Searchable Release Notes


chef-davin:
**Is your feature request related to a problem? Please describe.**
There is a frustration around not being able to search through release notes with the docs search feature.  Particularly for Chef Infra Client

In the previous format, to find things out about less-documented or undocumented things (target mode, powershell_exec, chef-run, compliance folders in cookbooks, etc...) I could go to the release notes and just search the giant page with the browser's search function, but now I have to try to remember the exact release version when a feature was introduced or improved and I can no longer find those things.

**Describe the solution you'd like**
It would be great if searching for an undocumented feature, I could at least be taken to the release page where that feature was introduced or updated so I could get some context around it.

**Describe alternatives you've considered**
Finding all the underdocumented things and creating Issues/PRs for each of them so they actually make it into the docs.

IanMadd:
I think this Issue can be lumped together with chef/chef-web-docs#3273.

I think the goal is that each product has a single page for release notes, that they're static pages and not loaded with JS which will allow search, and then each individual release has a separate page. So something like:

`https://docs.chef.io/release_notes_<PRODUCT>/` for a single static page,

and

`https://docs.chef.io/release_notes_<PRODUCT>/<VERSION>/` for static pages for each individual release note.

One goal is to maintain a single source for the RN so if they're updated in the S3 bucket, the updates will automatically show up in docs.chef.io. That could involve some kind of bash/Expeditor trickery or [Hugo's getJSON](https://gohugo.io/templates/data-templates/#get-remote-data).

🤔

IanMadd:
Close with #3623

Issue #3510, opened 2021-10-25
Promote date should be in the release note pages


tas50:
## General details

Link to page and section that needs updated: 

https://docs.chef.io/release_notes_server/?v=14.8.7

## Product

[ ] Chef Automate
[ ] Chef Habitat
[ ] Chef Infra Client
[ ] Chef Infra Server
[ ] Chef InSpec
[ ] All

## Describe the update that needs to be made. Be as specific as possible:

Customers need to be able to tell when a product was released from the release notes. This information needs to be included in these pages.

IanMadd:
From what I've found, the only source of release dates would be in the GitHub repos. We could generate a JSON file for each repo by looping through the API and get a list of dates by tag: 

For example, first here: https://api.github.com/repos/chef/chef/tags
Then here: https://api.github.com/repos/chef/chef/commits/a90cddca0af69895415cb34051b15966a9b28277

Then find a way to update this list at each future release with Expeditor. With that list we can add dates to the release notes pages.

tas50:
A better solution would probably be to extend omnitruck with information on the publish date. That information is most likely available in artifactory.

Issue #3273, opened 2021-07-12
Single page seamless release notes.


IanMadd:
**Is your feature request related to a problem? Please describe.**
Get seamless release notes on one page.

**Describe the solution you'd like**

There's a couple ways to handle this, one is to use the existing JavaScript to generate one long page, the other is to use Hugo's [GetJSON](https://gohugo.io/templates/data-templates/#get-remote-data) function to create a static page. I prefer the Hugo solution since Swiftype could index the content and make it searchable, also the content wouldn't reload every time the user navigates to a page. 

The Hugo solution would take longer to implement and the site would have to be rebuilt every time there's a new release, otherwise the new release notes content wouldn't get added to the site. That's probably already happening through the existing Expeditor config when we get content updates for each product release. I'd have to investigate this a bit.

One **big problem** with making a single page of release notes is that all the headings are set for multiple pages. Every heading would have to be bumped down automatically so the version number heading can be an H2 heading. 

**Describe alternatives you've considered**
We leave things as they are.

PR #3623, opened 2021-12-23
Static release notes

### LINKED ISSUE #3539: Release Notes Overhaul
## General details

People are looking for three features:

- **Searchable release notes**. This would have to be deployed as static files using Hugo and not JavaScript so that page or pages could be indexed with Swiftype. 
- **Single page release notes**. A single page of release notes so that the user could scroll through the entire page without having to click on each individual page if they're looking for a specific feature. This could be done in JavaScript. Ideally we could find a way to do this with Hugo.
- **Product release dates**. Add product release dates to the release note pages like in the [Automate release notes](https://docs.chef.io/release_notes_automate/?v=20210720135142).
- **Replace Showdown JS Markdown converter**. Showdown hasn't been updated in a few years and now we're getting security vulnerability warnings from dependabot.
## Review

Name of technical reviewer: Me

### Searchable Release Notes

Original Issue: chef/chef-web-docs#3519

The big issue here is that there is currently no way to get Hugo to pull in MD files from packages.chef.io and deploy them as a static page and Swiftype can't index text that's not static.

Options:
 - Hugo could grab Markdown text from a [JSON file](https://gohugo.io/templates/data-templates/#get-remote-data) from packages.chef.io and deploy it. We'd have to convert each MD file to JSON and anything that uses those files (downloads.chef.io) would have to be converted to handle JSON too.
 - Hugo could pull that content from another GitHub repository using the GitHub API and [Hugo's getJSON](https://gohugo.io/templates/data-templates/#get-remote-data).
 - Hugo could pull that content if we deployed the release note MD files from docs.chef.io from the static directory. Release notes could be released from docs.chef.io/data/release_notes/PRODUCT/VERSION.md. This would be the fastest option as far as Hugo build times are concerned.
 - Use **curl** to copy files before a build into static/data/release_notes/*. It would have to get the version numbers for each product, then loop through and grab each file. This could also be set to run only for a production deploy.
 - Someone is making a `getResource` function in Hugo which could grab MD files. Wait for them to finish which could be many months or years.
 
 
The downside of using Hugo's getJSON function is that it will add a considerable amount of time to the build process. I could set it so that Hugo only builds the release notes on a production deploy.

### Single Page Release Notes

Original Issue: chef/chef-web-docs#3273

This could be done in JavaScript. We could have one page for all release notes for a product and then another page with the existing style that lists a single page of release notes. 

### Product release dates

Original Issue: chef/chef-web-docs#3510

As far as I know there's no list or file of release note dates for products. It seems like there should be one somewhere though. The Automate release notes gets the release date by parsing the name of each MD file.

IanMadd:
## Description

Convert JS release notes to static release notes.

More details:

- removes JS release notes
- makes a single static release notes page for each product using Hugo's `resources.getRemote` function
- adds release notes for Habitat
- updates CSS for the TOC
- deploys Manage release notes using the same magic that we use for the other products

## Issues Resolved

#3539
#3273
#3519
#3621
#3047

## TODO

- [ ] ~~release note dates #3510~~
- [x] release notes for inspec cloud resources
    - [x] move release notes JSON to `assets` dir in inspec cloud resources and update release notes expeditor scripts
- [x] CSS

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


kagarmoe (approved):
Great work and also very clear. Thanks for removing the showdown mount, too. The speed at which the notes render is fantastic! Needs a rebase, but then you're OK to merge.

tenor-4121151

PR #3659, opened 2022-01-20
Remove ancient upgrade considerations

tas50:
This is totally useless at this point. It's mostly about the server upgrade actually and it's 11 -> 12.


kagarmoe:
I think the broad outline of the information could be useful to new Chefs. Conceptually, the information seems important in regard to managing node data and cookbook size. I don't think we explain it elsewhere.

How could we revise this section to be about "managing node data and cookbook size" and where could we relocate the content?

tas50:
None of this matters for a client upgrade though. Things like checking for name in metadata.rb were entirely for the Chef 11 -> 12 server upgrade. We have cookstyle rules for that. We also have cookstyle rules for binaries in the cookbook to keep size down. The recommended methods here of cleaning up old cookbooks are also wrong. This is the kind of doc that is worse than nothing. It takes people down a 2013 path of sadness.

kagarmoe (approved):
tenor-6178251

PR #3699, opened 2022-02-09
Add attribute query and delimiter details to ohai command line documentation

gene1wood:
## Description

This adds clarifying details to the ohai command line documentation on how to query for an attribute as well as how delimiters work in those attribute queries.

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


content/ctl_ohai.md
```diff
@@ -24,6 +24,30 @@ aliases = ["/ctl_ohai.html"]
 
 The following examples show how to use the Ohai command-line tool:
 
+### Query for a specific attribute
+
+To query for an attribute with the ohai command-line tool, pass the attribute
+as an argument
+
+```bash
+ohai os
```
tas50, line 32: It's probably worth mentioning what recipe style attributes these map to: `ohai os` being the same data as `node['os']`

gene1wood: Good idea, I've added that that in 43ffb7e0a78818dceffb3a192303c1206df48434

content/ctl_ohai.md
```diff
@@ -24,6 +24,31 @@ aliases = ["/ctl_ohai.html"]
...
+```
+
+Which would fetch the value found within Chef in `node[:os]` and return
```
tas50:
```suggestion
Which would fetch the value found within Chef Infra's node data at `node['os']` and return 
```

gene1wood: Good idea, that's better language

gene1wood: Since I merged this GitHub Suggested change from @tas50 the commit wasn't "Signed-Off by" and so the DCO CI check is now failing. Do I need to do anything to address that?

gene1wood:
The markdownlint issue is fixed in 1bf4f4ea2646f81fbacdcf51a6c2701404a7427e

IanMadd (approved):

PR #4134, opened 2023-06-28
Fix file resource properties

IanMadd:
## Description

A user pointed out that the file resource properties weren't correct. This should clean things up.

## Issues Resolved

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/data/infra/resources/file.yaml
+++ b/data/infra/resources/file.yaml
@@ -7,281 +7,351 @@ resources_common_notification: true
 resources_common_properties: true
 resource: file
 resource_description_list:
-- markdown: Use the **file** resource to manage files directly on a node.
-- note:
-    markdown: Use the **cookbook_file** resource to copy a file from a cookbook's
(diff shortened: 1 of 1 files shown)


data/infra/resources/file.yaml
```diff
@@ -7,320 +7,477 @@ resources_common_notification: true
...
+    path                       String # default value: 'name' unless specified
+    rights                     Integer, String
+    verify                     String, Block, Symbol
+    action                     Symbol # defaults to :create if not specified
+  end
 syntax_properties_list:
-- '`''/var/www/customers/public_html/index.php''` is path to the file and also the
-  filename to be managed'
-- '`content` defines the contents of the file'
+  - >-
```
johnmccrae, line 48: Ian is that '>' intentional? I'm asking in case this is part of markdown

PR #4267, opened 2024-04-24
Target Mode docs

IanMadd:
## Description

Add documentation about Target Mode.

## Issues Resolved

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


content/target_mode.md
```diff
@@ -0,0 +1,406 @@
...
+Target Mode has the following requirements:
+
+- A network-enabled system to execute Target Mode.
+- The `chef-client` CLI. This is included with Chef Workstation.
+- A credentials file which provides the system with information to connect to a target node.
+- A recipe that only includes Target Mode-enabled resources.
+
+## Credentials file
+
+The credentials file lists the nodes that Target Mode can connect to and the configuration for connecting to each node in TOML format.
```
IanMadd:
```suggestion
The credentials file defines the connection settings for each node in TOML format.
```

content/target_mode.md
```diff
@@ -0,0 +1,406 @@
...
+host = 'target.system.host.2.com'
+user = 'root'
+password = '123456'
+
+['192.168.0.252']
+host = '192.168.0.252'
+user = 'root'
+password = '123456'
+
+protocol = ssh
```
Vasu1105: This should be in single quotes. Otherwise it will give syntax error.

Vasu1105: Another thing which I want to keep as note here that protocol keyword does not work if using winrm as transport protocol. What works for both ssh and winrm is `transport_protocol`

content/target_mode.md
```diff
@@ -0,0 +1,406 @@
...
+Target Mode doesn't have a way to schedule Chef Infra Client runs on a node, but you can create a cron file that executes Target Mode on a regular schedule.
+
+For example, this create a cron file that executes Target Mode every thirty minutes:
+
+```ruby
+cat > /etc/cron.d/nodename.cron <<EOF
+PATH="/opt/chefdk/bin:/opt/chefdk/embedded/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+GEM_ROOT="/opt/chefdk/embedded/lib/ruby/gems/2.6.0"
+
+0,30 * * * * * chef-runner /usr/bin/chef-client -t <TARGET_NAME>
```
Vasu1105, line 365: As per my knowledge there is  missing peace of information here. 
1. Where this cron file should be? On node or on host?
2. Second when chef-client runs without --local-mode i.e -z option it will check for the client.pem file and client.rb file. That peace of information is missing here. Unless you have client.pem and client.rb file on the host the chef-client run will fail if not running in a local mode.

content/target_mode.md
rahulgoel1: As per Thomas H windows will not be supported for target mode. We will have to make that change later .  You can make the change mentioned above though. @tpowell-progress

content/target_mode.md
tpowell-progress: @Vasu1105 Given target mode, I would assume the host is the only one with chef-client, and the usual requirements for chef-client setup would apply. It might be worth calling that out specifically still.

PR #4278, opened 2024-06-25
Clarify required location of .d directories for each tool

jonlives:
## Description

The current documentation around the use of .d directories does not make it clear that different Chef tools require a specific directory name and will ignore any other .d directories not named correctly.

This PR clarifies the requirements of each tool.

## Check List

- [X] Spell Check
- [X] Local build
- [X] Examine the local build
- [X] All tests pass


IanMadd:
See changes merged in #4283

PR #4288, opened 2024-07-22
Install Script Changes for Omnitruck Documents

vviveksharma:
## Description

- As a part of this document change we have updated the direct install script URL to the operating system based install script API's that we have introduced.

## Issues Resolved

## Related PRs

https://github.com/inspec/inspec/pull/7118
https://github.com/chef/chef-server/pull/3869

## Check List

- [X] Spell Check
- [X] Local build
- [X] Examine the local build
- [X] All tests pass


IanMadd:
The InSpec and Server pages can't be edited here in the `_vendor` directory. Those are just dependency files, like a copy of Node package file. If those files are edited, they'll just be overwritten when any content is updated in the `_vendor` directory

The sources of those files are here:

- https://github.com/chef/chef-server/blob/main/docs-chef-io/content/server/upgrades.md
- https://github.com/inspec/inspec/blob/main/docs-chef-io/content/inspec/install.md

Once those files are edited and merged there, they can be updated in this repo.

karunesh-tech:
As per feedback from product - Sonia Arora(@sonarora) made changes in the draft documentation:

1/ Remove trial persona (only have commercial and community)
2/ Add the section to point to download portal
3/ Example of licenseId

karunesh-tech:
> The InSpec and Server pages can't be edited here in the `_vendor` directory. Those are just dependency files, like a copy of Node package file. If those files are edited, they'll just be overwritten when any content is updated in the `_vendor` directory
> 
> The sources of those files are here:
> 
> * https://github.com/chef/chef-server/blob/main/docs-chef-io/content/server/upgrades.md
> * https://github.com/inspec/inspec/blob/main/docs-chef-io/content/inspec/install.md
> 
> Once those files are edited and merged there, they can be updated in this repo.

@IanMadd 
Added changes as per above guidance:

https://github.com/inspec/inspec/pull/7118
https://github.com/chef/chef-server/pull/3869

content/install_chef_air_gap.md
```diff
@@ -161,9 +161,11 @@ custom template, and then modify the package and Ruby gem sources.
     template file name, be sure to replace `airgap.erb` with the
     template file you created during the last step.
 
-4. Update `airgap.erb` to replace `omnitruck.chef.io` with the URL of
+4. Update `airgap.erb` to replace `https://chefdownload-commericial.chef.io/install.sh?license_id=<LICENSE_ID>` with the URL of
     `install.sh` on your artifact store:
 
+    {{< note >}}Please follow steps mentioned in this document [Chef Software Install Script](/chef_install_script/). Please replace `<LICENSE_ID>` with your licenseId..{{< /note >}}
+
     ```ruby
```
IanMadd: I don't think this is correct. The `airgap.erb` file is generated from a template and unless that template has been updated (maybe it has?) the correct URL here is `omnitruck.chef.io`. This change  also doesn't make sense in this context since they're replacing the omnitruck URL with the URL of their own deployment.

@clintoncwolfe @sanghinitin

content/install_chef_air_gap.md
```diff
@@ -161,9 +161,11 @@ custom template, and then modify the package and Ruby gem sources.
     template file name, be sure to replace `airgap.erb` with the
     template file you created during the last step.
 
-4. Update `airgap.erb` to replace `omnitruck.chef.io` with the URL of
+4. Update `airgap.erb` to replace `https://chefdownload-commericial.chef.io/install.sh?license_id=<YOUR LICENSE ID>` with  the URL of
```
clintoncwolfe: Is this hostname correct, or does it contain a typo? It appears to misspell `commercial` as `commericial` (note extra I) This mistake is throughout this PR and https://github.com/inspec/inspec/pull/7118 . Please advise and correct if needed.

clintoncwolfe (dismissed):
I suspect there is a dangerous typo. Dismiss this review when resolved.

content/install_chef_air_gap.md
clintoncwolfe: It would seem in this context the user is looking for the omnitruck.chef.io value and inserting a per-org custom value. So the correct value would be `omnitruck.chef.io` i would think.

sandhi18: @clintoncwolfe,
Initially, the product install scripts were downloaded using the Omnitruck service only and therefore the value used was `omnitruck.chef.io`. But now, we have upgraded the Download APIs and added the support for the install scripts, and moving ahead these scripts will be available via the Download APIs only.
Therefore, we have added this change and updated the URLs from `omnitruck.chef.io` to the Downloads URL

Request you to please do suggest the correct value that should be used in this case

content/install_chef_air_gap.md
sandhi18: Yes, there was a typo error, and we have updated the URL to `chefdownload-commerical.chef.io`. 
Note: The hosted URL also has a typo (`commerical` instead of `commercial`) which cannot be modified

content/install_chef_air_gap.md
clintoncwolfe: @sandhi18 This is a different, special circumstance. Here, the customer is setting up an airgap environment with a customized download URL that only the customer will know. The value that will appear in the file that they need to replace is `omnitruck.chef.io` and they need to replace it with their customer value. You could say this:
```
Update `airgap.erb` to replace `https://chefdownload-commericial.chef.io/install.sh?license_id=<LICENSE_ID>` with the URL of
    `install.sh` on your artifact store:
    YOUR_CUSTOM_DOWNLOAD_URL
```

IanMadd: Yeah, but this assumes that the value in this airgab.erb file is `https://chefdownload-commericial.chef.io/install.sh?license_id=<LICENSE_ID>`, when it's actually `omnitruck.chef.io`. I ran the procedure described in this doc locally and it's `omnitruck.chef.io`, so unless someone has a fix in to update this, then this isn't going to be different. And if it is different, then it would likely be just the  FQDN `chefdownload-commericial.chef.io`.

sandhi18: @IanMadd @clintoncwolfe 
As per my understanding from the comment, it looks like the customer will be using some custom URLs so, as suggested by you, we have removed all the changes that were made regarding the install scripts in this particular file and moved it back to `omnitruck.chef.io`.
Request you to please review the changes and let me know if these look good
CC: @sonarora

karunesh-tech:
@IanMadd 
All the comments are taken care, please review and approve the PR and based on the release plan as per @sonarora you can proceed to merge the changes in order to make it live.
CC: @clintoncwolfe

IanMadd (approved):

PR #4373, opened 2025-01-28
update quiet mode documentation to reflect actual behavior

rishichawda:
## Description

This is related to the change that we did to fix quiet mode in https://github.com/chef/chef/pull/14779 (backported to chef 18 in https://github.com/chef/chef/pull/14817). 

In short, quiet mode wasn't working before. So with the fix we didn't want to break anyone's configuration. Hence, quiet mode was updated to be set to `false` by default. The documentation was also a little confusing since quiet mode is supposed to suppress the output completely vs "controlling verbosity".

## Issues Resolved

- https://progresssoftware.atlassian.net/browse/CHEF-9161
- https://progresssoftware.atlassian.net/browse/CHEF-20281

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass


IanMadd:
This change applies to Client 19 and 18.7.x.

IanMadd (approved):

PR #4419, opened 2025-04-28
Add information about public ca key, access key, and secret key.

IanMadd:
## Issues Resolved

- https://progresssoftware.atlassian.net/browse/CHEF-20190
- https://progresssoftware.atlassian.net/browse/CHEF-20219

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/360/saas/get_started/enroll_nodes.md
+++ b/content/360/saas/get_started/enroll_nodes.md
@@ -150,6 +150,74 @@ To configure the cookbooks and define enrollment settings, follow these steps:
     depends 'chef360-node-enroll', '~> 1.0.0'
     ```
 
+1. Create a copy of the Chef 360 SaaS public key and add it to the wrapper cookbook's `files` directory:
+
+    ```plaintext
+    -----BEGIN CERTIFICATE-----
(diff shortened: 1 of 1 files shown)


Stromweld:
The SaaS public key instructions here isn't a very good practice with CB writing we should be encouraging. Customers should have choice based on how they choose to approach it to get contents and store in a variable or attribute of wrapper cookbook and just passing it to the resource. I wouldn't recommend actually having any documentation on procedure as that should be documented in CB and we simply reference using the resource from the CB for registration. This way if it needs to change in the future we only have to update CB and not docs here.

content/360/saas/get_started/enroll_nodes.md
```diff
@@ -171,29 +230,30 @@ To configure the cookbooks and define enrollment settings, follow these steps:
     - `<CHEF_360_FQDN>` with the fully qualified domain name (FQDN) for your Chef 360 SaaS deployment.
     - `<ENROLLMENT_TYPE>` with either `full` or `partial` depending on the form of enrollment. Use `full` unless you must `partial`.
     - `<API_PORT>` with the API port configured in Chef 360 SaaS. The default value is `31000`.
```
cmailank: @IanMadd  API port should be 443 for SaaS

PR #4420, opened 2025-04-29
Obvious fix: Remove errant property value

williamtheaker:
Obvious fix

Looks like this unrelated string has been there for a few years oops.


IanMadd (approved):
Thanks @williamtheaker!

PR #4545, opened 2026-02-02
Update license key requirements for Client 19

IanMadd:
## Description

[Please describe what this change achieves]

## Issues Resolved

## Check List

- [ ] Spell Check
- [ ] Local build
- [ ] Examine the local build
- [ ] All tests pass

### DIFF
--- a/content/licensing/_index.md
+++ b/content/licensing/_index.md
@@ -33,7 +33,7 @@ The following products require a [Chef license key](license_key) to run, dependi
 For these products, you need a license depending on how you install it:
 
 - The Chef Habitat package for these products doesn't require a license key to run.
-- The Ruby gem requires a license key to run.
+- The Ruby gem downloaded from [RubyGems.org](https://rubygems.org/) requires a license key to run.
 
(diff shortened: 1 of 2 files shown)


content/licensing/_index.md
```diff
@@ -27,12 +27,13 @@ The following products require a [Chef license key](license_key) to run:
...
-- Chef InSpec 7
+- Chef InSpec 7 and later
+- Chef Infra Client 19 and later
 
 For these products, you need a license depending on how you install it:
 
-- The Chef Habitat package for Chef InSpec 7 doesn't require a license key to run.
-- The `inspec` gem requires a license key to run.
+- The Chef Habitat package for these products doesn't require a license key to run.
+- The Ruby gem requires a license key to run.
```
Stromweld: This should be updated to gem's downloaded from rubygems.org will require a license key during runtime to run.

We want to be careful because we are going to be offering our own rubygems download option that requires a license key to download but then doesn't require a license key at runtime. This could easily get confusing for customers.

PR #4711, opened 2026-08-03
Add unified resources catalog page at /resources

Kuxhul:
## Summary

Introduces a single, authoritative **Resources** discovery page at `docs.chef.io/resources` that lists every resource across **Chef Infra Client** and **Chef InSpec**, with a compatibility matrix, SEO-friendly descriptions, and links to each product's documentation.

Today, `/resources` is only a redirect to the Infra bundled pages, and resources are documented per product and per release with no unified view. This page provides that unified catalog **without replacing** the existing version-specific documentation — release pages continue to list only the resources introduced in each version.

## What's included

- **New page** `content/resources/_index.md` — top-level hub, registered in the Overview nav via page front matter (same pattern as `platforms.md` / `versions.md`; no `menu.toml` change).
- **New layout** `layouts/_default/resources_matrix.html` — adapted from the existing `infra_resources_all.html`. Merges the canonical Infra resource YAML (`data/client/19/resources`) with a committed InSpec snapshot into one **4-column compatibility matrix** (`Resource | Chef Infra Client | Chef InSpec | Description`). Checkmark cells link to each product's authoritative page.
- **New partial** `layouts/partials/resource_short_desc.html` — trims descriptions to the first sentence so cells stay scannable and SEO-friendly.
- **New data** `data/resources/inspec.yaml` — a committed snapshot of the InSpec resource catalog (name → platform, description, deprecated), generated once from `inspec/inspec`. InSpec docs are deployed as a separate site and aren't built from this repo, so a snapshot is the least-coupled way to include them.
- **`netlify.toml`** — documents the `/resources/*` redirect. It's non-forced (301), so Netlify serves the generated `/resources/index.html` for the hub while legacy per-resource deep links (for example `/resources/apt_package/`) still redirect to the Infra resource page.

## Data of record

Chef Infra Client resources: **181** · Chef InSpec resources: **407** · Supported by both: **16** · Total catalog: **572**.

## Design rationale

- **Reuse over new infrastructure.** The Infra half reads existing canonical YAML at build time (no duplication, auto-syncs). The new layout is a clone of the repo's existing all-resources generator.
- **No re-coupling of InSpec.** Re-importing InSpec as a Hugo module would drag ~407 pages and shortcodes into this build and undo the deliberate separate-site deployment. A thin committed snapshot mirrors the existing "resource data generated from a source repo, committed here" pattern.
- **No duplication of resource docs.** Rows link out to authoritative pages; the page is a discovery/index layer.

## Validation

- Isolated Hugo build against real data: no errors, no `raw HTML omitted` / `ZgotmplZ`.
- Matrix verified: 16 both-rows, 572 total rows, 181 Infra links, 407 InSpec links, no list bloat in cells.
- `markdownlint-cli2`: 0 issues. `vale --minAlertLevel warning`: 0 errors/warnings.
- `hugo_lint.sh` ASCII rule: new content and data files are ASCII-clean.

## Reviewer notes

- **Highest-attention item:** the `netlify.toml` redirect. Behavior relies on Netlify's file-precedence for non-forced redirects. Please confirm on a deploy preview that `/resources/` renders the new page and `/resources/apt_package/` still redirects.
- **InSpec snapshot** is a point-in-time capture; see the header comment in `data/resources/inspec.yaml` for the source and how to regenerate.

## Intentionally out of scope (follow-ups)

- A committed refresh script and CI wiring to auto-update the InSpec snapshot.
- Per-resource unified detail pages (currently link-out only).
- A filter/search UI on the matrix if the flat table proves large.


IanMadd:
I'm inclined to close this for these reasons:

1. This requires some level of maintenance as both Infra Client and InSpec add or remove resources. What is the process for updating the content in these data files? If it's not automatic, then expect that this resources page will drift as changes are made to InSpec and Client.

2. What's the value to end users? If you're developing a cookbook in Infra with Client resources and want to test your nodes, you'd have to navigate to another page to see what InSpec resources are available to test what you're developing in your cookbook. Also, does having one page that cross references resources across both products improve the user's ability to develop a cookbook?

3. I'm sure we have users running InSpec 5. and probably InSpec 6. We also have users running Client 18 and earlier as well. So this matrix doesn't support those users.

4. The /resources page was versioned and redirected to `client/19/resources/bundled/` not that long ago. So I expect that users have old bookmarks to /resources. It would be better to not add a page there and leave the redirect.

A better path is to add a "Matching InSpec resources" heading in the Infra resources that lists matching InSpec resources and supporting InSpec versions. This runs up against problem number one above, which is, how will this be maintained and verified.

An even better path is to improve the docs generation script in chef/chef so it's more accurate and requires less manual editing around each release and possibly improve the source resources code with better code examples and supporting text for the same reason.

IanMadd (changes requested):
See my comments

https://github.com/chef/chef-web-docs/pull/4711#issuecomment-5330495884

Kuxhul:
Thanks @IanMadd  for the review, this is really helpful.

1. The intent was to have this work similarly to our existing Supported platforms and Supported versions pages a single place to discover what resources Chef supports across Infra Client and InSpec. The version-specific resource docs would still remain the source of truth.

2. On the value proposition, I think your point about the cookbook authoring workflow is fair. Our thinking was mainly around breadth and discoverability, but the "Matching InSpec resources" idea might actually be more useful. I'm not tied to a particular approach here; I'd rather land on something that's genuinely useful.

3. You're also right about version coverage. A current-only view leaves out users on older InSpec and Client versions. We should cover those as well, and I'd be interested in your thoughts on the cleanest way to represent that

4. On the redirect, I agree we should be cautious about breaking existing /resources links. I'm happy to leave the redirect as-is and consider putting the new hub somewhere else if that's safer.

5. The maintenance point is probably the biggest consideration. We’re planning to automate the page update so that whenever a resource is added or removed, the page is updated automatically rather than relying on a manual snapshot. I like your suggestion of improving the generation in `chef/chef` and establishing a proper source of truth to support this.

Could we grab 15–20 minutes to discuss the maintenance and version coverage pieces? I think a quick conversation would be more productive than going back and forth in the comments. Really appreciate you taking the time to review this.

IanMadd:
I spent some time thinking about this and it just doesn't checkout:

- there isn't a 1:1 mapping between InSpec resources and Client resources. There are cases where a single InSpec resource could validate multiple Client resources and vice versa. It would give a false impression to state that specific Client resources map to specific InSpec resources.


A better path is to lay out a workflow for developing Client resources and validating that they work. Something like Create cookbook with Client resources --> Create InSpec profile that validates a client run --> Enable Compliance Phase --> Run Client --> Review results. And then create specific examples or a tutorial that shows exactly these things.

PR #4737, opened 2026-08-31
Fix broken Chef Automate download URL (m=x86_64 -> m=amd64)

RitvikPatil777:
## Summary

Fixes a broken Chef Automate CLI download command that has been reported by multiple customers (Tesco, Barclays) as failing this week.

## Root cause

The download command used throughout the Chef Automate docs was broken in three compounding ways:

1. **Wrong architecture parameter** — the URL used `m=x86_64`, which the download API doesn't recognize. Instead of returning a clear error, the API silently returns `HTTP 200` with an empty JSON body (`{"code":0,"status_text":"","message":""}`), which gets saved as if it were the real binary.
2. **Missing `-L` flag on `curl`** — the correct parameter (`m=amd64`) returns an `HTTP 302` redirect to the actual file. Without `-L`, `curl` doesn't follow the redirect and saves an empty file.
3. **Wrong output handling** — the API always returns a `.zip` archive, not a raw executable, but the docs treated the download as if it were already the finished `chef-automate` binary (`chmod +x` with no `unzip` step).

## Fix

- Changed `m=x86_64` → `m=amd64` in all Chef Automate download URLs.
- Added `-L` to `curl` commands that were missing it, so the redirect is followed.
- Changed output filename to `chef-automate.zip` and added an `unzip` step before `chmod +x`, so the extracted `chef-automate` binary is what actually gets marked executable.

## Verification

Tested live against the production download API using a valid trial license:

- Confirmed `m=x86_64` returns `HTTP 200` with a fake/empty JSON body (the broken behavior our customers hit).
- Confirmed `m=amd64` returns `HTTP 302` to a real `chef-automate_linux_amd64.zip`.
- Ran the exact fixed command as written in the docs end-to-end: downloaded the zip (matches reported file size), unzipped it, `chmod +x`'d it, and confirmed via `file` that the result is a valid `ELF 64-bit x86-64` executable.

## Files changed

- `content/automate/install.md`
- `content/automate/airgapped_installation.md`
- `content/automate/get_started.md`
- `content/automate/infra_server.md`
- `content/automate/ha_upgrade_introduction.md`
- `content/automate/migrate.md`
- `content/automate/restore.md`
- `content/automate/on_prem_builder.md`
- `content/automate/managed_services.md`
- `content/automate/major_upgrade.md`
- `content/automate/major_upgrade_4.x.md`
- `content/automate/ha_aws_deploy_steps.md`
- `content/automate/ha_aws_managed_deploy_steps.md`
- `content/automate/ha_on_premises_deployment_procedure.md`
- `content/automate/ha_chef_backend_to_automate_ha.md`

### DIFF
--- a/content/automate/airgapped_installation.md
+++ b/content/automate/airgapped_installation.md
@@ -29,7 +29,7 @@ To get a trial license for an airgapped host [contact Chef](https://www.chef.io/
 Download the Chef Automate command-line tool from the `current` [release channel]({{< relref "install.md#release-channels" >}}).
 
 ```shell
-curl "https://chefdownload-commercial.chef.io/stable/automate/download?p=linux&m=x86_64&license_id=<LICENSE_ID>" -o chef-automate && chmod +x chef-automate
+curl -L "https://chefdownload-commercial.chef.io/stable/automate/download?p=linux&m=amd64&license_id=<LICENSE_ID>" -o chef-automate.zip && unzip chef-automate.zip && chmod +x chef-automate
 ```
 
 Replace `<LICENSE_ID>` with your Progress Chef commercial license ID.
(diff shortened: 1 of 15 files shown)


Jayamathans12:
@punitmundra  @IanMadd  can you review this

IanMadd (approved):

vkarve-chef (approved):

punitmundra (approved):

Jayamathans12:
@IanMadd  can you merge this PR

